PR #2, opened 2015-01-13
"Unspike" (aka start over again)

henrikhodne:
This is very much still a WIP, but I figured it was time I made an "initial commit" and started to show some code for feedback.

Some of the things that I've done in this version "0.2" (or whichever version this is):

- Rename a bunch of things
  - "Worker" is now exclusively referring to this project as a whole/the process. Each "parallel worker" inside of a worker process is now called a "processor", and they all run in a single processor pool (so the relationship is something like this: we have one or more worker hosts, each running a single worker, each running a single processor pool, each running one or more processors, each running a single job (at a time)).
- The various "providers" have been extracted into a `backend` subpackage which could make it possible to add various custom providers in the future and makes it easier to have a very defined interface for them that does not depend on the internals of the worker.
- Be more conscious about errors and how to handle them.
- Be more conscious of the various features of the worker I didn't really keep in mind from the start on the first worker (such as graceful shutdown)

This code is far from complete and has many issues I am aware of and that I am working on, but hopefully you can see which direction I'm heading in.

Some TODOs:

- [ ] Write an integration test (and some unit tests where it makes sense). Unit tests are hard to write, since so many components are dependent on an external system such as RabbitMQ.
- [ ] Pass the `context.Context` around everywhere.
- [ ] Set up `SIGTERM` to cause a graceful shutdown.
- [ ] Send state updates to travis-hub over RabbitMQ
- [ ] Send log updates to travis-logs over RabbitMQ

/cc @meatballhat


meatballhat:
:heart_eyes_cat:

PR #16, opened 2015-06-11
Update usage of codegangsta/cli

meatballhat:


solarce:
I skimmed it, with the disclaimer of not being super proficient in Go yet, the changes make sense to me and I was able to follow them. I think that the docs updates seemed to cover everything in the code changes.

:+1:

PR #20, opened 2015-06-12
Move go sources back up top + symlink into ./src

meatballhat:
plus a few other tweaks to account for this change.


meatballhat:
NOPE NOPE NOPE

PR #25, opened 2015-06-23
Add packaging to builds

solarce:
- Builds packages using fpm
- Tests the installation of each package with docker
  - Using `dist: trusty`
- Uploads packages to packagecloud.io/travisci/worker/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ coverage.html
 .build/
 bin/travis-worker
 tmp/
+VERSION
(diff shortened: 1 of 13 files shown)


solarce:
- [] Need to add docs to README.md

solarce:
Add `docker` binary detection and "don't push to packagecloud unless it's building a new version" support to packaging scripts

solarce:
I think this is ready to merge now that we have a working set of default configs.

joshk:
Do it!

PR #26, opened 2015-07-07
docker: fix a panic by reversing a loop

henrikhodne:
The old loop seems to have gone from len(cpuSets) to 1, which seems like an off-by-one error (I think it should've gone from len(cpuSets)-1 to 0). I couldn't see a reason for the loop to go backwards, so I changed the loop to go from 0 to len(cpuSets)-1 in order to use the more standard Go loop.

/cc @meatballhat, who wrote the original loop (main question: any reason to go through the slice in reverse?)

### DIFF
--- a/src/github.com/travis-ci/worker/backend/docker.go
+++ b/src/github.com/travis-ci/worker/backend/docker.go
@@ -271,9 +271,9 @@ func (p *DockerProvider) checkoutCPUSets() (string, error) {
 
 	cpuSetsString := []string{}
 
-	for i := len(cpuSets); i > 0; i-- {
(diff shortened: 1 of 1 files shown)


solarce:
Worked in my testing

meatballhat:
FYI - I wrote the loop in reverse because @bcmills tol' me to! (kinda sorta)

:runner: :dash:

bcmills:
To be fair, I said that about "counting out N things" loops - for iterating over a map or slice a range-loop is clearly preferable.  =P

meatballhat:
@bcmills yay hi!  :wave:

PR #31, opened 2015-07-08
Fixing AWS detection logic for `docker --graph` path

solarce:


solarce:
Tested with vagrant and working

PR #33, opened 2015-07-08
Add support for lifecycle hook

henrikhodne:
This is needed before we can use this on EC2-Docker.


meatballhat:
Love it :+1: :heart_eyes_cat:

PR #38, opened 2015-07-09
Add GCE backend

meatballhat:
* [ ] add some danged tests
* [ ] does it blend?


joshk:
Feature idea: Allow for instances to be added to an instance group

meatballhat:
@joshk I'm curious; what does the instance group give us?

joshk:
@meatballhat nothing really, except a higher level overview of what we have running for each grouping of instance

meatballhat:
@joshk ah, OK, so how were you thinking they could be grouped?  By language?  staging vs. production?

joshk:
Staging / Production
Workers / Testing VMs
So four instance groups.

meatballhat:
@joshk :+1: love it

PR #41, opened 2015-07-10
Fix lint errors

henrikhodne:
This fixes all errors reported by `gometalinter` except:

- "Should have comment or be unexported"
- `errcheck` warnings on `defer`-ed calls


henrikhodne:
The reason why I skipped those is mainly that they take longer to fix.

henrikhodne:
(Please do review the `err` checking changes, some of them change behaviour while some only log errors)

meatballhat:
:sheep: :it:

PR #43, opened 2015-07-11
Fix more lint warnings/errors

henrikhodne:
Nothing in this pull request should change any behavior.

The largest change is probably the change in the `backend` package of making all implementations of `backend.Provider` and `backend.Instance` unexported. I did this as I found myself documenting all of the functions on these types as just `Start implements Provider.Start`, or something similar, and realised that the types don't really need to be exported at all, so they probably shouldn't.

My linter still reports some issues, but they are all either "cyclomatic complexity" issues, "duplicated code" issues or "error unchecked issues", all of which could only really be fixed by changing behavior, and not all of them _should_ be fixed, so I'm leaving that for another time.

(Also, this means that every exported type and function now has a documentation comment)

### DIFF
--- a/src/github.com/travis-ci/worker/config/config.go
+++ b/src/github.com/travis-ci/worker/config/config.go
@@ -49,9 +49,9 @@ type Config struct {
 	BuildCacheS3SecretAccessKey string
 }
 
-// ConfigFromCLIContext creates a Config using a cli.Context by pulling
-// configuration from the flags in the context.
-func ConfigFromCLIContext(c *cli.Context) *Config {
+// FromCLIContext creates a Config using a cli.Context by pulling configuration
+// from the flags in the context.
+func FromCLIContext(c *cli.Context) *Config {
 	cfg := &Config{
 		AmqpURI:       c.String("amqp-uri"),
 		PoolSize:      c.Int("pool-size"),
--- a/src/github.com/travis-ci/worker/config/config.go
+++ b/src/github.com/travis-ci/worker/config/config.go
@@ -126,7 +126,7 @@ func WriteEnvConfig(cfg *Config, out io.Writer) {
 
 	sortedCfgMapKeys := []string{}
 
-	for key, _ := range cfgMap {
+	for key := range cfgMap {
 		sortedCfgMapKeys = append(sortedCfgMapKeys, key)
 	}
 
--- a/src/github.com/travis-ci/worker/config/config.go
+++ b/src/github.com/travis-ci/worker/config/config.go
@@ -10,10 +10,6 @@ import (
 	"github.com/codegangsta/cli"
 )
 
-var (
-	zeroDuration time.Duration
-)
-
 // Config contains all the configuration needed to run the worker.
 type Config struct {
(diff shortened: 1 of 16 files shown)


src/github.com/travis-ci/worker/config/config.go
```diff
@@ -10,10 +10,6 @@ import (
 	"github.com/codegangsta/cli"
 )
 
-var (
-	zeroDuration time.Duration
```
henrikhodne: @meatballhat I'm guessing this was a leftover from some older version of the CLI parsing, but could you verify that this isn't needed anywhere? I did search through the codebase and couldn't find `zeroDuration` anywhere else.

meatballhat: It can be dropped now, yep!

src/github.com/travis-ci/worker/config/provider_config.go
```diff
@@ -9,19 +9,25 @@ import (
...
 }
 
+// GoString formats the ProviderConfig as valid Go syntax. This makes
+// ProviderConfig implement fmt.GoStringer.
 func (pc *ProviderConfig) GoString() string {
 	return fmt.Sprintf("&ProviderConfig{cfgMap: %#v}", pc.cfgMap)
 }
 
+// Map loops over all configuration settings and calls the given function with
+// the key and value. The settings are sorted so f i called with the keys in
```
meatballhat: "so f is called"

src/github.com/travis-ci/worker/config/provider_config.go
```diff
@@ -9,19 +9,25 @@ import (
...
+// alphabetical order.
 func (pc *ProviderConfig) Map(f func(string, string)) {
 	keys := []string{}
-	for key, _ := range pc.cfgMap {
+	for key := range pc.cfgMap {
```
meatballhat: I forgot that ranging over maps could be done like this :sweat_smile:

src/github.com/travis-ci/worker/processor_pool.go
```diff
@@ -46,6 +47,9 @@ func NewProcessorPool(hostname string, ctx gocontext.Context, hardTimeout time.D
 	}
 }
 
+// Each loops through all the processors in the pool and calls the given
+// function for each of them, passing in the index and the processor. The order
+// of the processors is the same for the same set of processors.
 func (p *ProcessorPool) Each(f func(int, *Processor)) {
```
meatballhat: `Each` or `Map`, @meatballhat??  I think `Each` is probably more appropriate because it's not like the aggregate results are being returned.

meatballhat:
:sheep: :it: :ship: :it: :shipit: LGTM !!!

src/github.com/travis-ci/worker/processor_pool.go
henrikhodne: `Map` to me is for a function that converts each element in a collection and returns a new collection, so I think `Each` is better here. I was a little confused about the similar function in `ProviderConfig` being named `Map` when it didn't return anything.

PR #45, opened 2015-07-11
metrics: add Memstats runtime metrics

henrikhodne:
This includes metrics on things like GC pause times. The metric calculation comes from http://pythonic.zoomquiet.io/data/20131112090955/index.html, but I'm not 100% sure I got everything translated correctly.


meatballhat:
Lovely!

PR #52, opened 2015-07-15
step/run-script: fix log timeout error message

### LINKED ISSUE #49: Bug in worker output
See this line:

https://github.com/travis-ci/worker/blob/eeef986dc3e10101e6f4685a4c4054afd5214f88/src/github.com/travis-ci/worker/step_run_script.go#L119

Shouldn't it print the log timeout instead of the hard time out?

In this example, the job is terminated after 10 minutes, as expected. But it prints 50m.

screen shot 2015-07-14 at 16 34 32

henrikhodne:
It was including the hard timeout instead of the log timeout. Since the log timeout is usually much shorter than the hard timeout, this could lead to some confusing error messages.

Fixes #49.

### DIFF
--- a/src/github.com/travis-ci/worker/step_run_script.go
+++ b/src/github.com/travis-ci/worker/step_run_script.go
@@ -116,7 +116,7 @@ func (s *stepRunScript) Run(state multistep.StateBag) multistep.StepAction {
 	case <-logWriter.Timeout():
 		cancelCtx()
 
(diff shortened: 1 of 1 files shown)


joshk:
Please let me know when a new deb package is released so I can update https://github.com/travis-ci/worker/issues/49

meatballhat:
:sheep: :it: LGTM

PR #54, opened 2015-07-15
Omit `uuid=` from log entries when empty

meatballhat:


henrikhodne:
Do we know _why_ the field is empty?

meatballhat:
@henrikhodne sorry, @joshk requested this and I didn't dig into why

henrikhodne:
I understand more what's going on now. It sounds like the UUID isn't set in the payload we get, so I think maybe we should change this in the place where we set the UUID instead so it doesn't get stored in the context at all?

henrikhodne:
LGTM :+1: :sparkles: :two_hearts: :tada:

PR #65, opened 2015-07-28
Adds pruning for packages

solarce:


utils/pkg/pkg_run
```diff
@@ -158,9 +162,22 @@ package_cloud_push () {
       export PC_REPO="worker-testing"
     fi
 
-    package_cloud push \
-      $PC_USER/$PC_REPO/$package_cloud_os/$platform_release \
-      tmp/output/$pkgtype/$platform_family/$platform_release/*.$pkgtype --verbose
+    PC_UPLOAD_FILE=$(ls $CHECKOUT_ROOT/tmp/output/$pkgtype/$platform_family/$platform_release/*.$pkgtype)
```
meatballhat: The @tianon in the back of my head says using `ls` here is asking for trouble :trollface:

meatballhat:
:ship: :it:

utils/pkg/pkg_run
solarce: What would you do instead? There can't be more than one file in the directory during a build because of how the script runs and by default it cleans out the folder at the end of the run, to account for running it locally

meatballhat: I think you can probably safely swap out `ls` with `echo` and let the shell do the expansion.  You might need/want to do some `shopt -s nullglob` so as not to end up with the unexpanded glob as your only entry.

tianon: Is the variable used anywhere besides the `echo` line immediately following it?

tianon: (if not, I'd just `echo whatever/*.glob` directly instead and let the shell do its thing)

solarce:
dancedancedance

PR #67, opened 2015-07-29
Local execution backend & job queue decoupling

meatballhat:
- [x] add a "file" job queue type which reads `.json` files from a directory, keeps some records maybe
- [x] does it blend?


henrikhodne:
LGTM :+1: :sparkles: :hearts:

PR #69, opened 2015-08-17
processor: check if we're shutting down before picking up job

henrikhodne:
If more than one case branch in a select statement can proceed, one is picked at random, which I think is the cause for graceful and forceful shutdown not to work a lot of times.


henrikhodne:
This should be tested in staging.

meatballhat:
I understand having the separate `select` prior to selecting on the incoming work channel, but what's the advantage of selecting over the shutdown channels twice?

meatballhat:
Discussed in meatspace.  :+1:

PR #75, opened 2015-08-19
Set grub cmdline args in grub.d/ file, account for local defaults file

meatballhat:


meatballhat:
Tested in GCE:

```
root@dan-worker-testing:~# cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-3.19.0-25-generic root=UUID=3fe10351-012c-4a61-9c3a-0e991108f2f5 ro cgroup_enable=memory s
wapaccount=1 apparmor=0 console=ttyS0
root@dan-worker-testing:~# apparmor_status
apparmor module is loaded.
apparmor filesystem is not mounted.
root@dan-worker-testing:~# grep /etc/default /etc/init/travis-worker.conf 
  if [ -f /etc/default/travis-enterprise ] ; then
    . /etc/default/travis-enterprise
  if [ -f /etc/default/travis-worker ] ; then
    . /etc/default/travis-worker
  if [ -f /etc/default/travis-worker-local ] ; then
    . /etc/default/travis-worker-local
```

PR #83, opened 2015-08-25
step/run-script: do not requeue job if max log length passed

henrikhodne:


henrikhodne:
Okay, this PR also includes some refactorings and fixes the bug where the hard timeout could cause a requeue too.

PR #84, opened 2015-08-27
Fix graceful shutdown, mark π

henrikhodne:
The `graceful` channel isn't buffered, so sending to it will block until
the processor is shut down. Calling `GracefulShutdown()` again will block
forever. Closing the channel will return immediately.

This should fix the issue with graceful shutdowns, since the processor pool waits for `GracefulShutdown()` to return before moving on to the next processor.


meatballhat:
:shipit: :sheep: :it: LGTM YISSS

PR #85, opened 2015-09-03
Image selector implementations and integration with GCE backend

meatballhat:


src/github.com/travis-ci/worker/image/api_selector.go
```diff
@@ -0,0 +1,123 @@
...
+type APISelector struct {
+	baseURL *url.URL
+}
+
+func NewAPISelector(u *url.URL) *APISelector {
+	return &APISelector{baseURL: u}
+}
+
+func (as *APISelector) Select(params *Params) (string, error) {
+	imageName := "default"
```
henrikhodne: Looks like this is only used in the `return` at the very end of the function. Can we remove this and just use a literal `"default"` at the end?

meatballhat: Heck yes

meatballhat:
This is currently working in staging: https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/435428

src/github.com/travis-ci/worker/backend/gce.go
```diff
@@ -379,7 +407,7 @@ func (p *gceProvider) Start(ctx gocontext.Context, startAttributes *StartAttribu
 		return nil, err
 	}
 
-	inst := p.buildInstance(startAttributes, scriptBuf.String(), image.SelfLink)
+	inst := p.buildInstance(startAttributes, image.SelfLink, scriptBuf.String())
```
meatballhat: _bugfix_

src/github.com/travis-ci/worker/backend/gce.go
```diff
@@ -644,6 +686,47 @@ func (p *gceProvider) imageForLanguage(language string) (*compute.Image, error)
...
+func (p *gceProvider) buildImageSelector() {
+	if p.imageSelector != nil {
+		return
+	}
+
+	switch p.imageSelectorType {
+	case "env":
+		p.imageSelector = image.NewEnvSelector(p.cfg)
+	case "api":
+		baseURL, _ := url.Parse(p.cfg.Get("IMAGE_SELECTOR_URL"))
```
henrikhodne: This ignores the error that can be returned if the URL can't be parsed. I think it's probably fine to `panic` (similar to `regexp.MustCompile` if it can't be parsed.

src/github.com/travis-ci/worker/backend/gce.go
```diff
@@ -644,6 +686,47 @@ func (p *gceProvider) imageForLanguage(language string) (*compute.Image, error)
 	return imagesByName[imageNames[len(imageNames)-1]], nil
 }
 
+func (p *gceProvider) imageForLanguage(language string) (*compute.Image, error) {
+	return p.imageByFilter(fmt.Sprintf(gceImageTravisCIPrefixFilter, language))
+}
+
+func (p *gceProvider) imageSelect(ctx gocontext.Context, startAttributes *StartAttributes) (*compute.Image, error) {
+	p.buildImageSelector()
```
henrikhodne: Could this cause a race condition if this method is called twice at the same time? There's a `if p.imageSelector != nil { return }` guard at the top of that method, but if the method is called twice very close to eachother they could both go past that?

henrikhodne: Forgot to add: One way that looks like it would resolve a potential race condition is moving this to the "initializer".

src/github.com/travis-ci/worker/image/api_selector.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (as *APISelector) queryWithTags(infra string, tags [][]string) (string, error) {
+	bodyLines := []string{}
+
+	for _, ts := range tags {
+		qs := url.Values{}
+		qs.Set("infra", infra)
+		qs.Set("fields[images]", "name")
+		qs.Set("limit", "1")
+		if len(tags) > 0 {
```
henrikhodne: I'm not quite sure I understand the logic here, but should this be `len(tags)` or `len(ts)`?

src/github.com/travis-ci/worker/image/api_selector.go
```diff
@@ -0,0 +1,181 @@
...
+			qs.Set("tags", strings.Join(ts, ","))
+		}
+
+		bodyLines = append(bodyLines, qs.Encode())
+	}
+
+	u, _ := url.Parse(as.baseURL.String())
```
henrikhodne: Another URL error being ignored, but if the error is caught on the base URL being parsed it _might_ not be necessary to catch it?

src/github.com/travis-ci/worker/image/api_selector.go
```diff
@@ -0,0 +1,181 @@
...
+	imageResp, err := as.makeImageRequest(u.String(), bodyLines)
+	if err != nil {
+		return "", err
+	}
+
+	if len(imageResp.Data) > 0 {
```
henrikhodne: This is more of a style thing, but I remember reading something somewhere about keeping the "happy path" at least indentation, so I'm wondering if it would be a good idea to invert this `if`, so that the `return "", nil` error condition is inside the braces and the `return imageResp.Data[0].Name, nil` is at the bottom of the method?

src/github.com/travis-ci/worker/image/env_selector.go
```diff
@@ -0,0 +1,146 @@
...
+			imageAliases[strings.ToLower(strings.Replace(key, "IMAGE_", "", -1))] = value
+		}
+	})
+
+	for _, aliasName := range aliasNamesSlice {
+		normalizedAliasName := strings.ToUpper(string(nonAlphaNumRegexp.ReplaceAll([]byte(aliasName), []byte("_"))))
+
+		key := fmt.Sprintf("IMAGE_ALIAS_%s", normalizedAliasName)
+		if !es.c.IsSet(key) {
+			// TODO: warn?
```
henrikhodne: We should probably do _something_ here, since the configuration is invalid. This happens during the worker "boot" process as far as I can tell, so might be as easy as just `panic`-ing?

src/github.com/travis-ci/worker/image/api_selector.go
meatballhat: :facepalm: yep. Thank you!

meatballhat: should be len(ts), that is

src/github.com/travis-ci/worker/backend/gce.go
meatballhat: Yep, def moving this to the init func.

src/github.com/travis-ci/worker/image/api_selector.go
meatballhat: :+1: heck ya

src/github.com/travis-ci/worker/image/env_selector.go
meatballhat: fixing!

PR #91, opened 2015-09-11
Post-GCE launch cleanup items

meatballhat:
including reinstalling all dependencies from nothing to ensure things are set up correctly wrt `gb`


henrikhodne:
LGTM :+1: :sparkles:.

As far as I can tell, the vendor update was mostly just the Docker library and its dependencies.

PR #93, opened 2015-09-22
backend/jupiter-brain: add support for image selectors

henrikhodne:
By default we're using the env image selector, since this is equivalent to the old selector, making this backwards compatible.


meatballhat:
LGTM :sheep: :it: !

PR #94, opened 2015-09-22
Attempt to fix cutoff logs

henrikhodne:
Included are a few changes to the Jupiter Brain log writer setup to attempt to fix the issue where the end of the logs aren't being sent to travis-logs. I'm not 100% confident that this will fix it, since these bugs would be in the GCE backend too, but the bug doesn't seem to be there from what I can see.


emdantrim:
:+1:

PR #96, opened 2015-10-27
docker: Add tests for Docker backend; Fix not passing HostConfig on /…

jacobgreenleaf:
…create call

The [Docker Remote API v1.20 /container/(id)/start](https://docs.docker.com/reference/api/docker_remote_api_v1.20/#start-a-container) call accepts a request body, although it is deprecated. The docker client the Travis worker uses encodes such a body, but fails to pass along the `HostConfig` that is used in the call to `/container/create`. As such, any option that is specified in the `HostConfig` fails to be preserved. 

In our case, this was the `Privileged` flag, which is important for running Docker-in-Docker.


jacobgreenleaf:
Mind re-running the build w/ tokens?

joshk:
Hi @jacobgreenleaf 

My name is Josh, I'm the Product Manager at Travis, and also work with Danish on some Enterprise bits.

Thanks for the PR. Just so I can make sure we are on the same page here, containers are started in privileged mode for Enterprise installations. This is done when creating the container, and before calling start. 

Docker in Docker is not currently support as it requires extra 'hacks' (https://github.com/jpetazzo/dind)

We are looking into this, but we don't have an ETA just yet sorry.

I'll talk further with our Platform team next week about our plans for Docker in Docker, and have @danishkhan circle back to you.

Thanks a bundle

Josh

jacobgreenleaf:
Hey Josh. 

> Thanks for the PR. Just so I can make sure we are on the same page here, containers are started in privileged mode for Enterprise installations. This is done when creating the container, and before calling start.

The bug here is not that you don't support privileged mode, it's that your support for it is broken by this bug which I fixed. 

- Jacob

meatballhat:
@jacobgreenleaf Thank you! :heart_eyes_cat: :metal:

PR #97, opened 2015-10-29
Rework API image selection tags

meatballhat:
such that the `language:<lang>` tags are gone, given limited usefulness, and a
new tag set with both `dist:` and `group:` is included in order to enable, e.g.:

``` yaml
language: ruby
group: edge
dist: precise
```


henrikhodne:
:+1: LGTM :heart: :sparkles:

PR #108, opened 2015-11-29
Image selection and start attribute default changes

meatballhat:
including:
- setting defaults for `language`, `dist`, `group`, and `os` for every job config
- setting `is_default=true` for image selection queries with a single param, in the rare case that `language` + `dist` + `group` matches nothing.


meatballhat:
Tested in org staging: https://staging.travis-ci.org/travis-repos/python-test-staging/builds/462581

PR #114, opened 2015-12-01
Tweaks to error messages and Sentry config

meatballhat:
- add bits to some log messages about what's happening next
- log the JSON when there are parsing errors
- reduce Sentry hook to panic and fatal levels by default, with option to include error level


henrikhodne:
The JSON payload often (always?) contains some credentials, so I'm not sure if logging it is a good thing.

meatballhat:
@henrikhodne :facepalm: of course.  Hmm.  My intent here is to get more information about the error.  Maybe I could do something a tiny bit clever and replace all strings with `"fafafaf"`.

henrikhodne:
I wonder if we can have scheduler put some kind of identifier in the headers of the message and then print that?

meatballhat:
That'd be cool.  I'll back out the JSON logging for now.

henrikhodne:
:sparkles: Looks good :+1: :blue_heart:

PR #117, opened 2015-12-02
docker: Add some documentation on how to disable resource allocation

henrikhodne:
I'm not 100% sure on the way this is documented, but couldn't find a better wording. Suggestions welcome!


meatballhat:
LGTM :sheep: :it:

joshk:
@danishkhan we might want to add this to the Enterprise docs as well

PR #118, opened 2015-12-03
Don't send zero timestamps in "finished" event to hub

henrikhodne:
The finish event can be sent without the received or started event having been sent if something errors in between. In that case, we would send 0001-01-01T00:00:00Z for received_at and/or started_at, causing hub to think the job had ran for roughly 2015 years.


meatballhat:
LGTM :shipit:

PR #124, opened 2015-12-08
gce: add support for a premium instance type

henrikhodne:
With this, the `vm_type` field on the payload we get from AMQP can be set to `premium` to get an `n1-standard-4` instead of `n1-standard-2`.


meatballhat:
LGTM :sheep: :it:

meatballhat:
@henrikhodne still look right?

PR #129, opened 2015-12-09
Switch to poll SSH instead of API for inst ready

meatballhat:
by using the script upload retries to do the job

- [x] also rip out the legacy image lookup &lt;/shame&gt;
- [x] test in staging

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/mitchellh/multistep"
 	"github.com/pborman/uuid"
 	"github.com/pkg/sftp"
 	"github.com/travis-ci/worker/config"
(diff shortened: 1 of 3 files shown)


solarce:
A skim of it and it seems good to me but I'd want a review from @henrikhodne before merging, as my Go is pretty minimal

backend/gce_test.go
```diff
@@ -176,6 +176,7 @@ func gceTestSetup(t *testing.T, cfg *config.ProviderConfig, resp *gceTestRespons
 		cfg = config.ProviderConfigFromMap(map[string]string{
 			"ACCOUNT_JSON": "{}",
 			"PROJECT_ID":   "project_id",
+			"IMAGE_ALIAS_": "default",
```
henrikhodne: This looks a little strange to me, what makes this necessary?

backend/gce.go
```diff
@@ -539,62 +468,131 @@ func (p *gceProvider) Start(ctx gocontext.Context, startAttributes *StartAttribu
...
+	zoneOpCall := p.client.ZoneOperations.Get(p.projectID, p.ic.Zone.Name, op.Name)
+
+	for {
+		newOp, err := zoneOpCall.Do()
+		if err != nil {
+			errChan <- err
+			return multistep.ActionHalt
+		}
+
+		if newOp.Status == "RUNNING" {
```
henrikhodne: Is it possible that an operation for some reason goes from running to done without us noticing, so this branch never triggers? Maybe it should be `newOp.Status == "RUNNING" || newOp.Status == "DONE"`? I'm not super-familiar with the GCE API, so I may be misunderstanding things, though.

backend/gce_test.go
meatballhat: It's totes strange, and specific to the tests, and means that language is `""` somewhere...

backend/gce.go
meatballhat: Heck ya totes true

backend/gce_test.go
meatballhat: found it

henrikhodne:
Makes sense to me. How many "boot polls" is a normal boot? (Are we likely to run out of retries if boots get "a little" slower?)

henrikhodne:
Other than that it looks good :sparkles: :+1: :heart:

meatballhat:
@henrikhodne My logic goes kind a like: `(upload retries) * ((length of upload attempt) + (post-fail sleep)) > (instance start timeout)`, so I don't *think* it's an issue (??)

henrikhodne:
> My logic goes kind a like: `(upload retries) * ((length of upload attempt) + (post-fail sleep)) > (instance start timeout)`, so I don't think it's an issue (??)

That looks right to me.

PR #133, opened 2015-12-10
update the default Docker adapter settings

joshk:
have the Docker adapter start privileged containers by default
this is required for using Docker in Docker for Enterprise, and also helps resolve certain issues reported by customers related to starting and using services within jobs.

Please note: I'm not sure if this was best to add to `/etc/default/travis-worker` or /etc/default/travis-enterprise`


solarce:
Merging this as-is would complicate things when we're ready to move EC2 to the Go Worker and use the packages for our AMIs, as we need privileged turned off in that context, for now.

Could we compromise on adding it _commented out_ and then adding an option to the installer to enable it? This would mean it's not hard to enable at install time and is also pretty painless to enable later.

meatballhat:
fwiw, proposed modification in #148 :heart_eyes_cat:

meatballhat:
Superseded, and merged!

meatballhat:
Egad!  I did this wrong.

PR #135, opened 2015-12-10
Add exponential backoff & optional skip to GCE instance stop

meatballhat:
to (theoretically) further reduce API usage

<del>**NOTE**: I'm not thrilled about this implementation</del>

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -79,7 +79,8 @@ var (
 		"ZONE":                  fmt.Sprintf("zone name (default %q)", defaultGCEZone),
 	}
 
-	errGCEMissingIPAddressError = fmt.Errorf("no IP address found")
+	errGCEMissingIPAddressError   = fmt.Errorf("no IP address found")
+	errGCEInstanceDeletionNotDone = fmt.Erronf("instance deletion not done")
 
 	gceStartupScript = template.Must(template.New("gce-startup").Parse(`#!/usr/bin/env bash
 {{ if .AutoImplode }}echo poweroff | at now + {{ .HardTimeoutMinutes }} minutes{{ end }}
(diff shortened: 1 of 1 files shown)


backend/gce.go
```diff
@@ -874,27 +875,42 @@ func (i *gceInstance) Stop(ctx gocontext.Context) error {
...
 					return
 				}
 
-				errChan <- nil
+				err = nil
 				return
 			}
 
-			time.Sleep(i.provider.bootPollSleep)
+			return
```
henrikhodne: This would just return from the function that's passed to `backoff.Retry`, I think the intention is to return from the "goroutine function"?

henrikhodne: Ignore that, I was misreading things.

backend/gce.go
```diff
@@ -868,42 +912,90 @@ func (i *gceInstance) RunScript(ctx gocontext.Context, output io.Writer) (*RunRe
...
-		return err
-	case <-ctx.Done():
-		return ctx.Err()
+		if newOp.Status == "DONE" && newOp.Error == nil {
+			err = nil
+			return
+		}
+
+		if newOp.Error != nil {
+			err = &gceOpError{Err: newOp.Error}
```
henrikhodne: I don't think this is correct. If we return an error from the function, it will just retry, but I think in this case the error is that the operation failed, and "retrying" will just give us the same error?

backend/gce.go
```diff
@@ -868,42 +912,90 @@ func (i *gceInstance) RunScript(ctx gocontext.Context, output io.Writer) (*RunRe
...
+			return
+		}
+
+		return
```
henrikhodne: We should probably return an error in this case, since we'd like it to retry if we get this far?

backend/gce.go
meatballhat: It *could* give us an error and then subsequently succeed, though, such as when we're hitting API rate limiting.

backend/gce.go
meatballhat: Agreed :facepalm:

backend/gce.go
henrikhodne: Wouldn't that return an error in the `err` from `zoneOpCall.Do()`, though? I thought `newOp.Error` being non-nil means that `newOp.Status` is DONE.

henrikhodne:
Looks good! :+1: :sparkles:

PR #140, opened 2015-12-11
gce: Add rate limiting

henrikhodne:
This rate limiter makes sure that we self-throttle on the number of API calls we make to the GCE API.


henrikhodne:
The algorithm is mostly borrowed from the wiki: https://github.com/golang/go/wiki/RateLimiting

PR #144, opened 2015-12-14
Allow config of script upload and startup timeouts

meatballhat:


henrikhodne:
Some of those argument lists are getting uncomfortably long, but looks good :+1: :sparkles: :heart:

PR #145, opened 2015-12-15
Generate SSH key pair in GCE

meatballhat:
rather than reading it as config


henrikhodne:
:sparkles: :heart_eyes_cat: :heart_eyes_cat: :sparkles:

PR #150, opened 2016-01-04
Package metadata touchup

meatballhat:
* mark config files as config files so that they are verified against in-place modifications and such
* fill in license, maintainer, and vendor fields


henrikhodne:
My version of `fpm` doesn't have `--default`, only `--deb-default`, at least according to the docs:

```
% fpm --help  | grep -- -default
    --[no-]deb-no-default-config-files (deb only) Do not add all files in /etc as configuration files by default for Debian packages. (default: false)
    --deb-default FILEPATH        (deb only) Add FILEPATH as /etc/default configuration
```

Should it be `--deb-default`, or is `--default` something else that's not in the `--help` output (or from a different version? I have fpm 1.4.0)

meatballhat:
@henrikhodne yep, it was a rotten guess on my part.  I was looking at the fpm source while writing those out the first time.  All better now.

meatballhat:
@henrikhodne the bit I was looking at when adding `--default` :smiley_cat: https://github.com/jordansissel/fpm/blob/15b050e5991389f7aabc0e04f7c8c52e90242a7b/lib/fpm/package/deb.rb#L146-L154

henrikhodne:
Ah, yeah, that is a bit confusing.

henrikhodne:
Looks good to me, although I'm not quite sure what the overlay and templates folders _do_.

meatballhat:
@henrikhodne those directories do nothing on their own, but instead are separate dirs for files that can be directly rsync'd (overlay) and those that are "templates", meant to satisfy specific `fpm` flags.

utils/package-build
```diff
@@ -63,27 +63,51 @@ __copy_binary() {
...
+}
+
+__run_fpm_rpm() {
+  __run_fpm
 }
 
 __run_fpm() {
-  fpm -f -s dir \
-    --log info \
+  local overlay_d="$CHECKOUT_ROOT/utils/package-overlay.d"
```
henrikhodne: This local variable doesn't appear to be used.

meatballhat: :+1:

PR #153, opened 2016-01-11
Unify configuration via cli.Context

meatballhat:
- removes (most) differences between top level and backend provider level configuration
- moves `--echo-config` :arrow_right: `echo-config`, `--list-backend-providers` :arrow_right: `list-backend-providers`, and adds `list-queue-types` command.


emdantrim:
makes sense to meeeeeeeeee

PR #155, opened 2016-01-19
Added the feature to support amqp to use custom TLS cert

chrismaul:
Adding the feature of adding a TLS cert. The cert should be configured as a PEM formatted string in the AmqpTlsCert config value.

Here are some documentation about this:
* https://godoc.org/github.com/streadway/amqp#hdr-Use_Case
* https://www.compose.io/articles/making-secure-connections-with-compose-rabbitmq/#finallygo

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -2,6 +2,7 @@ package worker
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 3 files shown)


chrismaul:
I am having trouble figuring out what I changed to make the build failed.

solarce:
Your failures were due to our build configs not being completely friendly to builds not in our repo.

If you can re-base with https://github.com/travis-ci/worker/pull/162 then your builds should not fail due to a lack of a packagecloud.io token

chrismaul:
The problem I was trying to solve should be fixed by https://github.com/travis-ci/worker/pull/160. Being able connect to a amqp with TLS with a custom TLS cert.

solarce:
I think the scenario of **ignore that you can't trust this cert** versus **here's how to trust this special cert** are different enough that we'd like support both.

We can certainly turn your changes into a PR from one of us, if you don't want to go through rebasing and so forth. What do you think?

chrismaul:
I can rebase it

chrismaul:
Rebased, and build passed. :-)

henrikhodne:
Thank you for submitting this! I merged this together with an extra option to set the path to a certificate in #164.

PR #156, opened 2016-01-20
Increase the log writer tick to 2 seconds

henrikhodne:


henrikhodne:
The idea behind this is that it'd put less strain on travis-logs by generating fewer log parts, while still being an acceptable log update interval for the "live" updates on the website.

solarce:
I apologize for the late comment, but I have some concerns over impact this might have user perception of _real timeness_ of logs and change in behavior for Enterprise users, if we just deploy it with the new default, straight away.

Some of my thoughts are
- can we make it configurable? It seems like a good idea since we've already encountered a scenario where we wanted to change it
- should we leave the default as 500ms and discuss why we want to increase our default configured setting in staging/production to 1s or 2s?
- is logs being strained atm and this will help or is this just a carryover from the logs issues back in mid-January 2016?

henrikhodne:
Reverted in #223 since after deploying this in a few places, we didn't observe any change in load on travis-logs, and there also seems to be quite a bit more headroom for travis-logs now than when this pull request was originally created.

PR #157, opened 2016-01-22
Add a Redis-backed rate limiter for GCE

henrikhodne:
This allows the rate limiter to share state across all workers, instead of just being local to a single worker process.

I tested this using a load testing script that ran three parallel processes, each spawning 50 goroutines that shared a rate limiter instance set to maximum 40 requests per second, and each goroutine continuously sending requests to a HTTP endpoint which tracked the maximum number of requests it got in a second. I let them run for about 15 minutes, and never did it go over 40 requests in a single second.

I also checked Redis to make sure the keys were actually cleaned up, and all of them expired and got removed correctly.

This could probably be improved in the future by using an implementation that allows you to spike over n req/s for a limited amount of time, as long as the average doesn't go over what you've set. That's probably a lot more complicated, and I don't think we need it to begin with.

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -457,7 +457,7 @@ func newGCEProvider(cfg *config.ProviderConfig) (Provider, error) {
 	}, nil
 }
 
-func (p *gceProvider) apiRateLimit() {
+func (p *gceProvider) apiRateLimit(ctx gocontext.Context) error {
 	metrics.Gauge("travis.worker.vm.provider.gce.rate-limit.queue", int64(p.rateLimitQueueDepth))
 	startWait := time.Now()
 	defer metrics.TimeSince("travis.worker.vm.provider.gce.rate-limit", startWait)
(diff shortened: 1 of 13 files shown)


backend/gce.go
```diff
@@ -364,18 +381,30 @@ func newGCEProvider(c ConfigGetter) (Provider, error) {
...
-	metrics.TimeSince("travis.worker.vm.provider.gce.rate-limit", startWait)
+	defer metrics.TimeSince("travis.worker.vm.provider.gce.rate-limit", startWait)
+
+	atomic.AddUint64(&p.rateLimitQueueDepth, 1)
 	// This decrements the counter, see the docs for atomic.AddUint64
-	atomic.AddUint64(&p.rateLimitQueueDepth, ^uint64(0))
+	defer atomic.AddUint64(&p.rateLimitQueueDepth, ^uint64(0))
+
+	for {
+		ok, err := p.rateLimiter.RateLimit("gce-api", p.rateLimitMaxCalls, p.RateLimitDuration)
```
henrikhodne: An error from the rate limiter is treated no differently than "rate limiter said 'no'" right now, is that what we want or should they be treated differently?

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,73 @@
...
+
+func NewNullRateLimiter() RateLimiter {
+	return nullRateLimiter{}
+}
+
+func (rl *redisRateLimiter) RateLimit(name string, maxCalls uint64, per time.Duration) (bool, error) {
+	rl.mutex.Lock()
+	defer rl.mutex.Unlock()
+
+	fmt.Printf("Rate limit %s, %d/%v\n", name, maxCalls, per)
```
henrikhodne: This debug statement snuck in at some point and should be removed.

PR #162, opened 2016-02-01
Only build packages when the API TOKEN is present as an environment variable

solarce:
Only build packages when the API TOKEN is present as an environment variable. This helps with external PRs building.


emdantrim:
:+1: :sheep: :it:

PR #165, opened 2016-02-04
GCE: Concurrency limiter

henrikhodne:
This implements a Redis-backed concurrency limiter that limits the number of instance start and stop operations that can be in progress at the same time.


henrikhodne:
Closing this, since we don't seem to hit this condition often (or at all?), and cloud-brain is taking over the responsibility for this.

PR #167, opened 2016-03-04
gce: allow looking for images in a different project

henrikhodne:


henrikhodne:
Tested on staging, and it worked as expected. :tada:

henrikhodne:
To make this work, you need to go into [Permissions](https://console.cloud.google.com/permissions/projectpermissions) on the project that contains the images, and add the GCE account that the worker uses with a Viewer role.

solarce:
What does this look like across projects? what kind of account is the worker in project using, that you can specify in the other?

henrikhodne:
The workers are generally using service accounts, so you specify the service account identifier (usually something like `username@project-name.iam.gserviceaccount.com`).

joshk:
THIS.
IS.
SO.
FRIGGIN.
AWESOME.
!.

PR #173, opened 2016-05-19
backend/gce: use hard timeout for the job instead of a global setting

henrikhodne:
It's possible that the job has an extended timeout, and if we are still using the global default for when the worker shuts down, it'll cause an infinite loop until the job is cancelled.

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -50,7 +50,6 @@ const (
 	defaultGCEStopPrePollSleep   = 15 * time.Second
 	defaultGCEUploadRetries      = uint64(120)
 	defaultGCEUploadRetrySleep   = 1 * time.Second
-	defaultGCEHardTimeoutMinutes = int64(130)
 	defaultGCEImageSelectorType  = "env"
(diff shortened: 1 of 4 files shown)


backend/gce.go
```diff
@@ -640,7 +634,12 @@ func (p *gceProvider) stepGetImage(c *gceStartContext) multistep.StepAction {
 
 func (p *gceProvider) stepRenderScript(c *gceStartContext) multistep.StepAction {
 	scriptBuf := bytes.Buffer{}
-	err := gceStartupScript.Execute(&scriptBuf, p.ic)
+	scriptData := gceStartupScriptData{
+		AutoImplode:        p.ic.AutoImplode,
+		HardTimeoutMinutes: int64(c.startAttributes.HardTimeout.Minutes()) + 10,
```
henrikhodne: Does +10 minutes make sense, or should it be more/less?

solarce: 10 seems sufficient to me.

PR #179, opened 2016-06-13
Add support for private IPs on GCE

igorwwwwwwwwwwwwwwwwwwww:
# todo

* [x] allow defining `Subnetwork` in which to launch instance


henrikhodne:
Would it make more sense to call the option "public IP"? All instances have a private IP, this seems to be toggling whether or not to also have a public IP.

igorwwwwwwwwwwwwwwwwwwww:
@henrikhodne done! I also went ahead and updated all deps (because i needed the latest version of google compute).

PR #183, opened 2016-06-29
Toggle graceful shutdown and pause on SIGWINCH

meatballhat:


.gitignore
```diff
@@ -47,3 +47,4 @@ vendor/**
 /*-success
 .crossdeps
 build/
+.write-test
```
meatballhat: I added this after noticing it was left behind when running `travis-worker --queue-type file`.

processor_pool.go
```diff
@@ -107,10 +108,23 @@ func (p *ProcessorPool) Run(poolSize int, queue JobQueue) error {
 
 // GracefulShutdown causes each processor in the pool to start its graceful
 // shutdown.
-func (p *ProcessorPool) GracefulShutdown() {
+func (p *ProcessorPool) GracefulShutdown(pause bool) {
```
meatballhat: Hm, should this be named `togglePause` instead of `pause`?

processor_pool.go
```diff
@@ -107,10 +108,23 @@ func (p *ProcessorPool) Run(poolSize int, queue JobQueue) error {
...
+	log := context.LoggerFromContext(p.Context)
+
+	if pause {
+		if p.paused {
+			log.Info("finishing wait group to unpause")
+			p.processorsWG.Done()
+		} else {
+			log.Info("incrementing wait group for pause")
+			p.processorsWG.Add(1)
+			p.paused = true
```
henrikhodne: This and the read above it should probably be wrapped in a mutex to prevent race conditions. You'd have to send two signals really close to eachother for a race condition to happen, but if it does happen it looks to me like you'd either (1) get a panic because the wait group goes negative or (2) graceful shutdown stops working because each signal "pauses" and adds 1 to the wait group (so +2 in total), but when you unpause, it only removes 1, so you'll always have 1 left in the wait group.

meatballhat:
@henrikhodne The mutex alone wasn't preventing panics, hence the conversion from bool to counter.

processor_pool.go
```diff
@@ -30,6 +30,8 @@ type ProcessorPool struct {
 	processorsLock sync.Mutex
 	processors     []*Processor
 	processorsWG   sync.WaitGroup
+	pauseCount     int
+	pauseLock      *sync.Mutex
```
henrikhodne: Is there a reason why this is a pointer and not just `sync.Mutex`?

henrikhodne:
I now see that the entire `GracefulShutdown` method has a lock around it (to access the `processors` slice, so the extra pause lock might not be needed (and introduces the possibility of introducing deadlocks later if the pause lock is used in other places, although that's not necessarily a reason to remove the extra lock, more something to be aware of when working on this code later).

meatballhat:
@henrikhodne given that the extra lock didn't prevent a panic, I'm leaning toward removing it

henrikhodne:
:+1: from me on removing it. I'm fairly certain it doesn't provide any extra safety due to the lock at the top of the function.

processor_pool.go
meatballhat: Wat.  The methods take pointer receivers.  😐 https://godoc.org/sync#Mutex  ah whatteber, ripping it out anyway 😸

henrikhodne:
LGTM :+1:

I think there's a possibility of a race condition still if `GracefulShutdown` is called in parallel, but it's only called from the signal handler and the signal handler only handles one signal at a time, so it should be fine.

meatballhat:
@henrikhodne How would you prevent that race?

meatballhat:
@henrikhodne and also, if such a race happened, would it matter given that only a single "toggle" is allowed?

henrikhodne:
As far as I can see, if two calls to `GracefulShutdown` was called, it could _potentially_ do both of the `p.pauseCount++`s before the `if` section starts, which means that one of the two could happen:

- `p.processorsWG.Done()` is called twice without `p.processorsWG.Add(1)` being called at all, causing a panic once the second to last processor finishes (since the wait group then will go negative).
- `p.processorsWG.Done()` isn't called at all, both calls just cause the `return` branch to happen, and the worker is stuck.

Also, I just realized _yet again_ that `GracefulShutdown` has a lock at the top, so it's not possible to cause it to call in parallel, so we are all good. I keep forgetting that it's there…

PR #186, opened 2016-07-13
amqp-job-queue: don't log the channel when close fails

henrikhodne:
The channel object is very long and doesn't contain any exported information that has helped in debugging in the past, so let's remove it to make the log messages easier to read.


meatballhat:
😻

PR #189, opened 2016-07-15
gce: print last error in upload loop if it times out

henrikhodne:


henrikhodne:
Gonna merge and deploy this on Monday.

PR #192, opened 2016-08-12
Add a timeout for shutdown after AMQP error

henrikhodne:
After an AMQP error we should be able to shut down quite fast, we just need to wait for any running instances to shut down. If we don't shut down quickly, that probably means that something has gotten stuck, and we should force shut down.


cli.go
```diff
@@ -361,5 +361,8 @@ func (i *CLI) amqpErrorWatcher(amqpConn *amqp.Connection) {
 	if ok {
 		i.logger.WithField("err", err).Error("amqp connection errored, terminating")
 		i.cancel()
+		time.Sleep(2 * time.Minute)
```
henrikhodne: Is two minutes too long? We only really need to wait for the API calls to GCE/vSphere/etc. to go through to shut down running instances, there's nothing to send back to AMQP, and if the API calls to shut down instances fails we have cleaner scripts that should fix it.

cli.go
```diff
@@ -361,5 +361,8 @@ func (i *CLI) amqpErrorWatcher(amqpConn *amqp.Connection) {
...
+		i.logger.Error("timed out waiting for shutdown after amqp connection error")
+		os.Exit(1)
```
henrikhodne: Is `os.Exit` good or should we `panic`?

emdantrim: I think just exit is good, maybe with a call to sentry so that we can get some easy metrics on how often this happens.

cli.go
meatballhat: LGTM :shipit:

cli.go
```diff
@@ -361,5 +361,8 @@ func (i *CLI) amqpErrorWatcher(amqpConn *amqp.Connection) {
 	if ok {
 		i.logger.WithField("err", err).Error("amqp connection errored, terminating")
 		i.cancel()
+		time.Sleep(2 * time.Minute)
+		i.logger.Error("timed out waiting for shutdown after amqp connection error")
```
meatballhat: Nobody commented on this line yet so here

solarce: i approve of this comment

cli.go
emdantrim: 2 might be too long, I just was saying that because it was a little underneath half the alert threshold for missing queue consumers.

cli.go
solarce: I approve of this comment approved

cli.go
henrikhodne: We could `logger.Fatal`, and then `logrus` will terminate the process, and the sentry hook will send to Sentry.

emdantrim: sounds as good as anything to meee

cli.go
henrikhodne: Reducing it to one minute instead, that should be plenty I think.

henrikhodne:
Ready for review.

henrikhodne:
> Would it make sense to dump the stacks of all goroutines so that we can effectively debug what got stuck? Because honestly it sounds to me like there is some goroutine still running somewhere and it is not hooked up to the cancellation properly.

That's a good idea, I'll change it to panic instead of just `os.Exit(1)`.

PR #193, opened 2016-08-12
Expect Docker 1.12.0 instead of 1.9.1 to be preinstalled

henrikhodne:
The build is currently failing (I think) because the preinstalled Docker version is now 1.12.0 instead of 1.9.1. This PR just updates what version it expects, I'm hoping that's all that's needed.

@meatballhat or @solarce Do you know if this is correct, or if we specifically want to test against 1.9.1?


meatballhat:
:sheep: :it:

PR #195, opened 2016-08-25
backend/jupiterbrain: add context to all errors

henrikhodne:
This uses github.com/pkg/errors a lot to provide some context for all errors, which makes it easier for us to figure out exactly where an error comes from when we see an error in the logs, instead of just "file not found" or "EOF".

### DIFF
--- a/vendor/manifest
+++ b/vendor/manifest
@@ -295,8 +295,8 @@
 			"importpath": "github.com/pkg/errors",
 			"repository": "https://github.com/pkg/errors",
 			"vcs": "git",
-			"revision": "a2d6902c6d2a2f194eb3fb474981ab7867c81505",
-			"branch": "master",
+			"revision": "17b591df37844cde689f4d5813e5cea0927d8dd2",
+			"branch": "HEAD",
 			"notests": true
 		},
 		{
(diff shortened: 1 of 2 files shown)


meatballhat:
Is there an update for `vendor/manifest` or were we already using this package?

Otherwise :sheep: :it:

henrikhodne:
> Is there an update for vendor/manifest or were we already using this package?

We were already using it, but I just pushed an upgrade to update to 0.7.1 (and thereby also locking it to a released version instead of some in-between SHA).

vendor/manifest
```diff
@@ -295,8 +295,8 @@
 			"importpath": "github.com/pkg/errors",
 			"repository": "https://github.com/pkg/errors",
 			"vcs": "git",
-			"revision": "a2d6902c6d2a2f194eb3fb474981ab7867c81505",
-			"branch": "master",
+			"revision": "17b591df37844cde689f4d5813e5cea0927d8dd2",
+			"branch": "HEAD",
```
meatballhat: the most exciting branch of all

meatballhat:
:shipit:

PR #196, opened 2016-08-28
Bump to go 1.7, use native cross-compilation, ensure static

meatballhat:


meatballhat:
Currently trying to convince the static to be really binary, which I'm assuming is because some dependency somewhere is using `cgo` in such a way that the go toolchain is refusing to respect `-extldflags "static"`.

meatballhat:
Successfully in use by https://github.com/travis-ci/packer-templates/pull/276 :tada:

PR #199, opened 2016-09-01
AMQP consumer leak fix

henrikhodne:
This fixes a bug that was mostly visible when using `SIGTTOU` to scale down a worker as it was running, but it would also be a problem during graceful shutdown. When a processor is told to shut down (or not gracefully, for that matter), it would finish its current job and then stop processing more jobs. However, the AMQP consumer would still be active, and AMQP would deliver a job to it, expecting worker to work on it. This would lead to a job being "stuck" in the worker, but nothing in the worker would've logged that it was received, and it would probably remain there until the worker either restarted or lost AMQP connection (causing the connection to close and all consumers to get cancelled).


henrikhodne:
This PR has quite a few whitespace changes, so use https://github.com/travis-ci/worker/pull/199/files?w=1 to ignore those in the diff

henrikhodne:
For some reason I can't reproduce the errors locally at all (even though they are definitely errors) :/

meatballhat:
:trumpet: :clap:

henrikhodne:
Yay, tests are passing now.

@meatballhat This is ready for review whenever you have a moment.

PR #202, opened 2016-09-14
Implement heartbeat handling

meatballhat:
for use with cyclist/whatever remote supervisor-like relationship

### DIFF
--- a/backend/cloudbrain_client.go
+++ b/backend/cloudbrain_client.go
@@ -8,7 +8,7 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/cenkalti/backoff"
+	"github.com/cenk/backoff"
 	"github.com/pkg/errors"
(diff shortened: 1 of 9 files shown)


Lyoness:
LGTM :+1: 
what do you mean 
> whatever remote supervisor-like relationship

meatballhat:
By this:

> whatever remote supervisor-like relationship

I mean that this "heartbeat" implementation is really a two-fer, as it both informs the thing receiving the HTTP requests that the worker making the request is "alive", and the worker making the request uses the response to decide whether it should shut down gracefully.  Surely, this could be made more direct/clean/normal, but this implementation has been working well in the [legacy travis-worker](https://github.com/travis-ci/travis-worker) deployment for over 18mo, so I'd like to stick with it :smile_cat:.

PR #204, opened 2016-09-17
Hook and heartbeat updates

meatballhat:
to work moar better with cyclist and terraform config

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -198,7 +198,7 @@ func (i *CLI) handleStartHook() {
 	}
 
 	i.logger.WithField("start_hook", hookValue).Info("running")
-	outErr, err := exec.Command(hookValue).CombinedOutput()
+	outErr, err := exec.Command("sh", "-c", hookValue).CombinedOutput()
 	if err == nil {
 		return
 	}
@@ -217,7 +217,7 @@ func (i *CLI) handleStopHook() {
 	}
 
 	i.logger.WithField("stop_hook", hookValue).Info("adding deferred execution")
-	outErr, err := exec.Command(hookValue).CombinedOutput()
+	outErr, err := exec.Command("sh", "-c", hookValue).CombinedOutput()
 	if err == nil {
 		return
 	}
--- a/cli.go
+++ b/cli.go
@@ -204,8 +204,9 @@ func (i *CLI) handleStartHook() {
 	}
 
 	i.logger.WithFields(logrus.Fields{
-		"err":     err,
-		"out_err": outErr,
+		"err":        err,
+		"output":     outErr,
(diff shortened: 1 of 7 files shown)


cli.go
```diff
@@ -198,6 +191,48 @@ func (i *CLI) Run() {
...
+	}).Error("start hook failed")
+}
+
+func (i *CLI) handleStopHook() {
+	hookValue := i.c.String("stop-hook")
+	if hookValue == "" {
+		return
+	}
+
+	i.logger.WithField("stop_hook", hookValue).Info("adding deferred execution")
```
henrikhodne: This line doesn't seem correct anymore. It used to be called before we did the `defer`, but now it's run "inside" of the `defer`-ed function.

cli.go
```diff
@@ -198,6 +191,48 @@ func (i *CLI) Run() {
...
+
+	parts := stringSplitSpace(hookValue)
+	outErr, err := exec.Command(parts[0], parts[1:]...).CombinedOutput()
```
henrikhodne: Are there any multi-word parameters to the commands we use (for example `/bin/sh -c "foo bar"`)? If so, this won't work, but if everything is single-word it should work fine.

meatballhat: I'm not planning to add anything multi-word, but yeap it's definitely not foolproof.

cli.go
```diff
@@ -198,6 +190,68 @@ func (i *CLI) Run() {
...
+	go i.heartbeatHandler(hbURL, strings.TrimSpace(hbTok))
+}
+
+func (i *CLI) handleStartHook() {
+	hookValue := i.c.String("start-hook")
+	if hookValue == "" {
+		return
+	}
+
+	i.logger.WithField("start_hook", hookValue).Info("running")
```
henrikhodne: Would a better message be "running hook" or "running start hook"? I'm thinking if we want to search for these log messages, searching for `msg="running start hook"` might be nicer than `msg="running" start_hook=`?

cli.go
```diff
@@ -198,6 +190,68 @@ func (i *CLI) Run() {
...
+	}).Error("start hook failed")
+}
+
+func (i *CLI) handleStopHook() {
+	hookValue := i.c.String("stop-hook")
+	if hookValue == "" {
+		return
+	}
+
+	i.logger.WithField("stop_hook", hookValue).Info("running")
```
henrikhodne: Similarly, should this be "running hook" or "running stop hook" instead?

henrikhodne:
LGTM now :+1: Left two more questions above, but I'll leave it up to you if you want to change that or not.

PR #207, opened 2016-09-22
backend/jb: print more info on boot timeouts

henrikhodne:


henrikhodne:
This probably won't work, since the goroutine is probably gonna be stuck in `net.Dial` and not return in time.

backend/jupiterbrain.go
```diff
@@ -339,7 +367,12 @@ func (p *jupiterBrainProvider) Start(ctx gocontext.Context, startAttributes *Sta
 		}
 		instance.Stop(ctx)
 
-		return nil, ctx.Err()
+		select {
+		case err := <-errChan:
+			return nil, err
+		case <-time.After(2 * time.Second):
```
henrikhodne: This can possibly be reduced. `net.Dial` will return when the context is cancelled, so we don't need to wait for a full second. The HTTP call to Jupiter Brain might still be stalling, though.

backend/jupiterbrain.go
```diff
@@ -286,11 +299,19 @@ func (p *jupiterBrainProvider) Start(ctx gocontext.Context, startAttributes *Sta
...
+				case <-time.After(p.bootPollSleep):
+				case <-ctx.Done():
+				}
+
 				continue
 			}
 
-			conn, err := net.Dial("tcp", fmt.Sprintf("%s:22", ip.String()))
+			dialer := &net.Dialer{
+				Timeout: time.Second,
```
henrikhodne: Is one second enough? Too high? Too low?

meatballhat: When I'm in doubt, I prefer to push it to a configurable value :grin:

That being said, one second timeout on dial seems low to me.  `3s`?  idk

backend/jupiterbrain.go
meatballhat: Again with pushing such values to config.  I realize this means more overhead in the form of config parsing, struct members, and argument passing.  I'd much rather have the option to alter a runtime config than have to wait for a PR->build->binary cycle.

PR #210, opened 2016-09-22
Revert "attempt to log upstream error on start failure"

henrikhodne:
The previous change caused the error to be logged twice, and breaks the
"constant" msg field.

This reverts commit 4ce7d2cfc42e536043277e8257823dd2c7cd32b6.


henrikhodne:
@meatballhat r?

henrikhodne:
@igorwwwwwwwwwwwwwwwwwwww Do you have any thoughts on this since you made the original change?

igorwwwwwwwwwwwwwwwwwwww:
yep, I believe this should be fine. it seemed like it was missing from the logs when i made the change, but I think I was mistaken at the time 👍

PR #214, opened 2016-10-12
backend/jupiterbrain: parse SSH key on backend init

### LINKED ISSUE #206: Extract SSH private key parsing to only do it once per process
Right now we load and parse the SSH key multiple times per job (once per SSH connection), and this is the largest source of CPU usage for the worker currently.

I've started to draft out this locally, will open a PR :soon:, just opening this to track it.

henrikhodne:
Instead of loading, parsing and decrypting the key on every SSH connect, this does it only once when the worker is first set up.

Gonna test this in staging once the build passes and I have a binary to test with.

Fixes #206.

### DIFF
--- a/backend/cloudbrain.go
+++ b/backend/cloudbrain.go
@@ -13,13 +13,12 @@ import (
 	"github.com/Sirupsen/logrus"
 	"github.com/mitchellh/multistep"
 	"github.com/pkg/errors"
-	"github.com/pkg/sftp"
 	"github.com/travis-ci/worker/config"
 	"github.com/travis-ci/worker/context"
 	"github.com/travis-ci/worker/image"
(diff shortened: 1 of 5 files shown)


henrikhodne:
Tested in staging by running [a few jobs](https://staging.travis-ci.org/henrikhodne-test/travis-test-staging/builds/513937) in parallel, and I observed no issues.

henrikhodne:
On second thought, it'd probably be best to port this over to the other providers too.

henrikhodne:
Will make some more changes to this tomorrow to port the changes over to the other backends too.

henrikhodne:
Alright, I ported everything over to the other backends by extracting the functionality to an `ssh` subpackage.

PR #222, opened 2016-10-24
Delete job from job-board on finish

igorwwwwwwwwwwwwwwwwwwww:
Worked on this with @Lyoness during the offsite. The pass-down of `jobBoardURL` and a few other values still needs to be implemented. We should also check the http response code.


http_job.go
```diff
@@ -100,6 +100,26 @@ func (j *httpJob) currentState() string {
...
+	// From: ${UNIQUE_ID}
+
+	// copy jobBoardURL
+	url := *q.jobBoardURL
+	url.Path = "/jobs/" + string(id)
+	url.Userinfo = nil
+
+	client := &http.Client{}
+
+	req, err := http.NewRequest("DELETE", url.String(), nil)
```
henrikhodne: This error needs to be checked.

http_job.go
```diff
@@ -100,6 +100,26 @@ func (j *httpJob) currentState() string {
...
+
+	req.Header.Add("Travis-Site", q.site)
+	req.Header.Add("Authorization", "Bearer "+j.Payload().JWT)
+	req.Header.Add("From", q.uniqueID)
+
+	resp, err := client.Do(req)
```
henrikhodne: This error should also be checked.

PR #228, opened 2016-11-09
amqp-job: implement state message counter

henrikhodne:
This counter helps Hub reorder the messages if they arrive in a different order than Worker sent them in, which sometimes causes a "requeue" immediately after "received" to be interpreted as "requeue" and _then_ "receive", causing the job to get stuck.

/cc @svenfuchs, is this roughly what you thought in terms of JSON payloads? I.e. `state_update_count` in the top-level of both the payload from Scheduler and the payload to Hub?

/cc @solarce

### DIFF
--- a/amqp_job.go
+++ b/amqp_job.go
@@ -22,6 +22,7 @@ type amqpJob struct {
 	startAttributes *backend.StartAttributes
 	received        time.Time
 	started         time.Time
+	stateCount      uint
 }
 
 func (j *amqpJob) GoString() string {
@@ -126,9 +127,13 @@ func (j *amqpJob) sendStateUpdate(event, state string) error {
 	}
 	defer amqpChan.Close()
 
+	j.stateCount++
 	body := map[string]interface{}{
 		"id":    j.Payload().Job.ID,
 		"state": state,
+		"meta": map[string]interface{}{
+			"state_update_count": j.stateCount,
+		},
 	}
 
 	if j.Payload().Job.QueuedAt != nil {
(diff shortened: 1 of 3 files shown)


svenfuchs:
@henrikhodne super, so glad to see this happening! ❤️ 

I have just recently started including a "meta" hash into messages to Hub, e.g. [here] (https://github.com/travis-ci/travis-gatekeeper/blob/92b0f18ce72c621ff56464a2737d51305483b25c/lib/travis/gatekeeper/service/build/cancel.rb#L34).

Would that be a too annoying in Go to do in this case as well?

henrikhodne:
@svenfuchs So, something like this?

``` Go
err := j.sendStateUpdate("job:test:reset", map[string]interface{}{
    "id":    j.Payload().Job.ID,
    "state": "reset",
    "meta": map[string]interface{}{
        "state_update_count": j.stateCount,
    },
})
```

I.e. the following JSON:

``` JSON
{
    "id": 1234,
    "state": "reset",
    "meta": {
        "state_update_count": 1
    }
}
```

svenfuchs:
@henrikhodne looks good to me! wdyt?

henrikhodne:
Updated to pass the counter inside a meta block. Should test this out in staging first to make sure hub ignores the value for now, then I think this is good to merge.

svenfuchs:
@henrikhodne is the input format the same? i.e. is Scheduler supposed to include the same meta hash?

henrikhodne:
> is the input format the same? i.e. is Scheduler supposed to include the same meta hash?

Ah, no, right now it just expects `state_update_count` at the top level of the object, I'll push it into a meta object.

henrikhodne:
Tested this on MacStadium staging and hub seemed to deal with the new fields just fine without any errors.

Issue #184, opened 2016-06-30
Add metrics to count jobs finished.


emdantrim:
After some discussion, we decided it would be really neat to have the worker output a metric which counts jobs completed, which could give further insight into how many jobs we process at one time, so that we can observe our overall performance on the platform. 

Preferably these would be completed jobs tagged with build images used, build infrastructure used, and "errored"/"failed"/"passed", which could also help us in the future, as we push out new build images, to see how many jobs error or fail when we push out a new build image. 

The sum of these would also be useful for determining the overall performance of a build infrastructure -- seeing how quickly we're processing jobs, since we can't really see how quickly we're processing jobs from a queue which is having jobs added to it (which is our queue, in 99% of cases.)

meatballhat:
I'm pretty sure this is covered now with #231 and the preexisting per-processor metrics.  WDYT?

emdantrim:
This is closely related, yes, but I think that my original request here was more to get a count of overall job throughput, so that we can see if there is a spike in errored builds, or an overall unexpected trend to process fewer builds over time -- more from a job performance perspective.

Seems to me like #231 is more of a measure of per-process reliability.

meatballhat:
That's def true regarding #231.  I can see how sending mark-based metrics for FinishState would allow us to alert on things like errored jobs crossing a threshold.

henrikhodne:
Appears solved in #232, closing.

PR #232, opened 2016-11-17
Add a meter for job finish state

meatballhat:
Related to #184


meatballhat:
Ah, I didn't realize that was newer.  Fixing...

PR #237, opened 2016-12-01
Some Jupiter Brain backend refactoring

henrikhodne:
Mainly:

- Split the "wait for SSH to come up" loop into two separate functions"
- Move HTTP API things to a separate type

The general goal here was to make the functions smaller and easier to read (they used to be bigger than what could fit on my screen, which I think is too big for it to be easy to keep track of what's going on).


henrikhodne:
Pushed a bug fix that adds a sleep between the creation and the polling.

PR #242, opened 2016-12-12
check job-board response status code on image selection

igorwwwwwwwwwwwwwwwwwwww:


image/api_selector.go
```diff
@@ -102,16 +103,27 @@ func (as *APISelector) makeImageRequest(urlString string, bodyLines []string) (*
...
 		defer resp.Body.Close()
+
 		responseBody, err = ioutil.ReadAll(resp.Body)
-		return
+		if err != nil {
+			return err
+		}
+
+		if resp.StatusCode != 200 {
+			return errors.Errorf("expected 200 status code, received status=%d body=%q",
```
henrikhodne: Would it be a good idea to include from where we expected the 200 status code? Something like "expected 200 status code from Job Board, received…".

emdantrim: don't want to be pedantic, but why not `job-board` instead of `Job Board`?

igorwwwwwwwwwwwwwwwwwwww:
If job-board has any kind of intermittent error that causes it to 500 we would probably want to retry. e.g. what if the dyno crashes or such.

image/api_selector.go
igorwwwwwwwwwwwwwwwwwwww: updated!

igorwwwwwwwwwwwwwwwwwwww:
PR is now passing! 🎈

PR #248, opened 2017-01-09
check tagsets for trailing commas before querying job-board

igorwwwwwwwwwwwwwwwwwwww:
When .travis.yml contains trailing commas, those commas propagate all
the way through to to worker, which builds them into tags and passes
those tags on to job-board.

job-board does not like tags with commas and returns a 500, which then
causes a retry loop.

This patch adds a clearer error message for when that happens, displays
this error message to the user, and also makes sure we do not requeue.

### DIFF
--- a/image/api_selector.go
+++ b/image/api_selector.go
@@ -239,7 +239,7 @@ func (as *APISelector) buildCandidateTags(params *Params) ([]*tagSet, error) {
 	for _, ts := range result {
 		for _, tag := range ts.Tags {
 			if strings.Contains(tag, ",") {
-				return result, workererrors.NewWrappedJobAbortError(errors.Errorf("job was aborted because tag %v contained \",\", this can happen when .travis.yml has a trailing comma", tag))
+				return result, workererrors.NewWrappedJobAbortError(errors.Errorf("job was aborted because tag \"%v\" contained \",\", this can happen when .travis.yml has a trailing comma", tag))
 			}
 		}
 	}
--- a/image/api_selector.go
+++ b/image/api_selector.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/cenk/backoff"
 	"github.com/pkg/errors"
-	workerErrors "github.com/travis-ci/worker/errors"
+	workererrors "github.com/travis-ci/worker/errors"
 )
 
 const (
@@ -239,7 +239,7 @@ func (as *APISelector) buildCandidateTags(params *Params) ([]*tagSet, error) {
 	for _, ts := range result {
 		for _, tag := range ts.Tags {
 			if strings.Contains(tag, ",") {
-				return result, workerErrors.NewWrappedJobAbortError(errors.Errorf("tag %v contained \",\", which is not supported by job-board -- check .travis.yml for trailing comma", tag))
+				return result, workererrors.NewWrappedJobAbortError(errors.Errorf("tag %v contained \",\", which is not supported by job-board -- check .travis.yml for trailing comma", tag))
 			}
 		}
 	}
--- a/image/api_selector.go
+++ b/image/api_selector.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/cenk/backoff"
 	"github.com/pkg/errors"
+	workerErrors "github.com/travis-ci/worker/errors"
 )
 
 const (
@@ -35,7 +36,12 @@ func NewAPISelector(u *url.URL) *APISelector {
 }
 
 func (as *APISelector) Select(params *Params) (string, error) {
-	imageName, err := as.queryWithTags(params.Infra, as.buildCandidateTags(params))
+	tagSets, err := as.buildCandidateTags(params)
+	if err != nil {
+		return "default", err
+	}
+
+	imageName, err := as.queryWithTags(params.Infra, tagSets)
 	if err != nil {
 		return "default", err
 	}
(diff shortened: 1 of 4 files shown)


step_start_instance.go
```diff
@@ -31,6 +33,14 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 		context.LoggerFromContext(ctx).WithField("err", err).Error("couldn't start instance")
 		context.CaptureError(ctx, err)
 
+		jobAbortErr, ok := errors.Cause(err).(workerErrors.JobAbortError)
+		if ok {
+			logWriter := state.Get("logWriter").(LogWriter)
+			logWriter.WriteAndClose([]byte(jobAbortErr.UserFacingErrorMessage()))
+
+			return multistep.ActionHalt
```
henrikhodne: I think you need to set the state to errored with `buildJob.Finish(FinishStateErrored)` as well, since `stepUpdateState` is "after" `stepStartInstance` in the processor stack, and as such won't get run if we halt here.

image/api_selector_test.go
```diff
@@ -204,13 +204,36 @@ func TestAPISelector_SelectDefaultWhenBadJSON(t *testing.T) {
...
+	actual, err := as.Select(&Params{
+		Infra:    "test,",
+		Language: "ruby,",
+		OsxImage: "meow,",
+		Dist:     "yosamitty,",
+		Group:    "dev,",
+		OS:       "osx,",
+	})
+	assert.Equal(t, actual, "default")
+	assert.Error(t, err, "tag \"a\" contained \",\", which is not supported by job-board -- check .travis.yml for trailing comma")
```
igorwwwwwwwwwwwwwwwwwwww: this test case should be failing, gonna have to dig into why it isn't tomorrow

image/api_selector.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	"github.com/cenk/backoff"
 	"github.com/pkg/errors"
+	workerErrors "github.com/travis-ci/worker/errors"
```
meatballhat: Convention is to import as smashedlowercase, e.g. `workererrors`.  I don't recall if there's a lint check for this, though, and I'm assuming not...

image/api_selector.go
```diff
@@ -230,7 +236,15 @@ func (as *APISelector) buildCandidateTags(params *Params) []*tagSet {
 		sort.Strings(ts.Tags)
 	}
 
-	return result
+	for _, ts := range result {
+		for _, tag := range ts.Tags {
+			if strings.Contains(tag, ",") {
+				return result, workerErrors.NewWrappedJobAbortError(errors.Errorf("tag %v contained \",\", which is not supported by job-board -- check .travis.yml for trailing comma", tag))
```
meatballhat: Is the intent that this error propagates up to the job output log?  If so, I think that this error can/should also be handled at a higher level prior to scheduling, such as at the gatekeeper level, so that an error message can be provided at the requests view and can report with richer context than "tag %v".  Even if this error string remains, I don't think it should contain any reference to `job-board`, or really to `.travis.yml`, but instead should be fairly low-level, optimized for backend debugging.

igorwwwwwwwwwwwwwwwwwwww: I definitely agree that this error should be caught at the gatekeeper level. But I do want to also handle it in worker as a fallback. If you think that is overkill/redundant, I'd be glad to hear your thoughts.

I'll see if I can make the error better, and potentially split it into a user facing one and a log-facing one.

henrikhodne: I think this specific error should probably be caught in Gatekeeper for a nicer user-facing error, but I also don't think there should be a case where invalid input data into worker can cause an infinite requeue loop, so some form of error handling should be done in worker.

image/api_selector.go
igorwwwwwwwwwwwwwwwwwwww: fixed!

step_start_instance.go
igorwwwwwwwwwwwwwwwwwwww: good catch! fixed :)

image/api_selector.go
igorwwwwwwwwwwwwwwwwwwww: error message has been updated

image/api_selector_test.go
henrikhodne: The "message" part of `assert.Error` (the third param) is the message that is printed if `err` is `nil`, not something that is matched against the error message itself. So as long as `err` is non-nil, then `assert.Error` isn't going to fail the test.

henrikhodne: You may be looking for [`assert.EqualError`](https://godoc.org/github.com/stretchr/testify/assert#EqualError).

igorwwwwwwwwwwwwwwwwwwww: that explains it! thank you <3

image/api_selector.go
```diff
@@ -230,7 +236,15 @@ func (as *APISelector) buildCandidateTags(params *Params) []*tagSet {
 		sort.Strings(ts.Tags)
 	}
 
-	return result
+	for _, ts := range result {
+		for _, tag := range ts.Tags {
+			if strings.Contains(tag, ",") {
+				return result, workererrors.NewWrappedJobAbortError(errors.Errorf("job was aborted because tag \"%v\" contained \",\", this can happen when .travis.yml has a trailing comma", tag))
```
meatballhat: Is this message intended for end users or Travis humans?

igorwwwwwwwwwwwwwwwwwwww: It's intended for travis humans -- though it will potentially show up to end users until we have better handling earlier in the chain.

meatballhat: Does this error have to bubble up to the output log?  If it can be kept internal-only, then I'd say we could include even more context, if available.

igorwwwwwwwwwwwwwwwwwwww: It does bubble up to the output log. I wanted to at least include a hint for users when this happens.

igorwwwwwwwwwwwwwwwwwwww:
I am going to test this on staging today.

PR #251, opened 2017-01-11
track queue time via queued_at field from payload

igorwwwwwwwwwwwwwwwwwwww:
This patch will give us a metric of the actual queue wait time experienced by users.

Depends on https://github.com/travis-ci/travis-scheduler/pull/55

### DIFF
--- a/job_test.go
+++ b/job_test.go
@@ -70,7 +70,7 @@ var jsonPayload = `
 }
 `
 
-func TestUnarshalJobPayload(t *testing.T) {
+func TestUnmarshalJobPayload(t *testing.T) {
 	var job JobPayload
 
 	err := json.Unmarshal([]byte(jsonPayload), &job)
--- a/job_test.go
+++ b/job_test.go
@@ -1,0 +1,81 @@
+package worker
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var jsonPayload = `
+{
+	"type": "test",
+	"vm_type": "default",
+	"queue": "builds.docker",
+	"config": {
+		"language": "rust",
+		"rust": "stable",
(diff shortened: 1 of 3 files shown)


henrikhodne:
Note that `queued_at` is reset in a requeue, so this won't reflect longer times when requeues are happening, so the reported numbers will probably be too low. Probably better than nothing, though.

igorwwwwwwwwwwwwwwwwwwww:
Good point, that is something I'd want to track separately though. I think with queue wait + requeue we already have a pretty good idea of how long things are taking.

It's also worth noting that because we overwrite that information it is currently very difficult to even get that kind of number. AFAIK the only way at the moment is to look through log files.

henrikhodne:
Would it be a good idea to split up the metric per queue so we can tell what the queue time for each individual queue is?

igorwwwwwwwwwwwwwwwwwwww:
@henrikhodne imo all metrics should be split per queue -- if that is not the case yet, we should change that. wdyt?

henrikhodne:
> imo all metrics should be split per queue -- if that is not the case yet, we should change that. wdyt?

Come to think of it they are split per worker hostname in the source, so we can do some globbing to make charts that are split per worker name and do it that way. So ignore what I said, I forgot we already do this :P

job_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"ssh_key": null,
+	"timeouts": {
+		"hard_limit": null,
+		"log_silence": null
+	},
+	"cache_settings": {}
+}
+`
+
+func TestUnarshalJobPayload(t *testing.T) {
```
meatballhat: `s/Unarshal/Unmarshal/`

meatballhat: uh whoops I meant to hit "Request Changes" :woman_shrugging:

PR #253, opened 2017-01-16
prepare CHANGELOG for 2.6.0 release

igorwwwwwwwwwwwwwwwwwwww:


CHANGELOG.md
```diff
@@ -1,4 +1,29 @@
...
+## [2.6.0] - 2017-01-16
+### Added
+- cli: log processor pool total on shutdown
+- amqp_job: meter for job finish state (#184)
+- amqp_job: track queue time via queued_at field from payload
+- amqp_job_queue: log job id immediately after json decode
+- capture every requeue and send the error to sentry
+
+### Changed
+- update to latest go version
```
henrikhodne: This seems to be a "build system" change, which I'm not sure belong in the change log?

CHANGELOG.md
```diff
@@ -1,4 +1,29 @@
 ## Unreleased
+
+## [2.6.0] - 2017-01-16
+### Added
```
henrikhodne: Should we include that the `REGION` and `SUBNETWORK` config flags to `backend/gce` are now documented?

CHANGELOG.md
```diff
@@ -1,4 +1,29 @@
...
+- cli: log processor pool total on shutdown
+- amqp_job: meter for job finish state (#184)
+- amqp_job: track queue time via queued_at field from payload
+- amqp_job_queue: log job id immediately after json decode
+- capture every requeue and send the error to sentry
+
+### Changed
```
henrikhodne: Should we include that the SSH code was consolidated to be shared by all the backends? In theory it should be a no-op, but there were minor differences in how each of the backends SSH-ed, so there might be some minor behavior differences.

CHANGELOG.md
igorwwwwwwwwwwwwwwwwwwww: I mainly wanted to mention it because it's security relevant

CHANGELOG.md
igorwwwwwwwwwwwwwwwwwwww: I can add it if you like, but when I put together the changelog I skipped over changes that _should not_ affect behaviour.

CHANGELOG.md
meatballhat: Maybe move it to a `Security` section?

From http://keepachangelog.com/en/0.3.0/:
> `Security` to invite users to upgrade in case of vulnerabilities.

igorwwwwwwwwwwwwwwwwwwww: alright, will do :)

PR #255, opened 2017-01-19
processor: open log writer before startInstance step, so it can write logs to the user

igorwwwwwwwwwwwwwwwwwwww:
We are currently experiencing panic()s because stepStartInstance may try to
write a message out to the user, but the log writer has not been initialized
yet.

This patch moves the opening of the log writer further back in time.

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -189,10 +189,7 @@ func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
 			generator: p.generator,
 		},
 		&stepSendReceived{},
-		&stepOpenLogWriter{
-			logTimeout:   logTimeout,
-			maxLogLength: p.maxLogLength,
-		},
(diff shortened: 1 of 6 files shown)


henrikhodne:
What do we normally set the log timeout to again? I think if the instance boot takes longer than the log timeout this might potentially cause issues, but I don't remember what the timeout is so it might not be an issue. I think the longest boot timeout we have right now is 5 minutes.

igorwwwwwwwwwwwwwwwwwwww:
Good point. The [default](https://github.com/travis-ci/worker/blob/master/config/config.go#L25) is 10m and I was not able to find any overrides in keychain.

igorwwwwwwwwwwwwwwwwwwww:
TODO: move logWriter.SetTimeout() call to separate step that gets run just before stepRunScript.

PR #260, opened 2017-01-20
backend/docker: Add support for specifying size of /dev/shm

meatballhat:
for all containers launched by the backend.  Note that this *does not* allow for per-job shm sizing.


backend/docker.go
```diff
@@ -29,6 +29,7 @@ var (
 		"CERT_PATH":       "directory where ca.pem, cert.pem, and key.pem are located (default \"\")",
 		"CMD":             "command (CMD) to run when creating containers (default \"/sbin/init\")",
 		"MEMORY":          "memory to allocate to each container (0 disables allocation, default \"4G\")",
+		"SHM":             "/dev/shm to allocate to each container (0 disables allocation, default \"64MB\")",
```
henrikhodne: `humanize.ParseBytes` parses MB as 1000 * 1000 bytes, not 1024 * 1024 bytes, so if you say `SHM=64MB` you'll get something other than the default (64 * 1024 * 1024). I think 1024-based makes sense for memory things, so maybe change this string to say `default \"64MiB\"` instead?

meatballhat: There seems to be confusion/inconsistency about this in the docker source and docs, too, where it is claimed that the default is `"64MB"`, but the numeric value is `67108864` (`1024 * 1024 * 64`).  Docker appears to support (via https://github.com/docker/go-units) `MB`, `M`, and `m`, but *not* `MiB`:

``` bash
$ docker version --format='{{.Client.Version}}'
1.12.0
$ docker run --rm -it --shm-size=64MB alpine:latest df -h /dev/shm
Filesystem                Size      Used Available Use% Mounted on
shm                      64.0M         0     64.0M   0% /dev/shm
$ docker run --rm -it --shm-size=64M alpine:latest df -h /dev/shm
Filesystem                Size      Used Available Use% Mounted on
shm                      64.0M         0     64.0M   0% /dev/shm
$ docker run --rm -it --shm-size=64m alpine:latest df -h /dev/shm
Filesystem                Size      Used Available Use% Mounted on
shm                      64.0M         0     64.0M   0% /dev/shm
$ docker run --rm -it --shm-size=64MiB alpine:latest df -h /dev/shm
docker: invalid size: '64MiB'.
See 'docker run --help'.
```

How about we switch to the slightly more ambiguous yet more consistent with Docker `m`/`M`?

henrikhodne: What I mean is this code:

``` Go
shm := uint64(1024 * 1024 * 64)
if cfg.IsSet("SHM") {
	if parsedShm, err := humanize.ParseBytes(cfg.Get("SHM")); err == nil {
		shm = parsedShm
	}
}
```

We're passing in a byte count to docker, not a string. Docker may parse MB as 1024-based, but `humanize.ParseBytes` does not, so I think either changing the doc string to "64 MiB" or changing the default value to `64 * 1000 * 1000` would be best (probably the former, since memory things are usually 1024-based).

PR #267, opened 2017-01-31
backend/gce: log upload failure with level=error

igorwwwwwwwwwwwwwwwwwwww:
This should help the observability of this issue: https://github.com/travis-pro/team-blue/issues/618


igorwwwwwwwwwwwwwwwwwwww:
Ah, figured out why it is info. Because the actual error in this case is the timeout, so it can be misleading. I think I'll close this ticket for now.

PR #269, opened 2017-02-01
Add steps for initial sleep and interstitial cancellation checks

henrikhodne:
We currently have a backlog of "stale" jobs and we'd like to avoid booting an instance if we know it's going to be cancelled immediately.


igorwwwwwwwwwwwwwwwwwwww:
We should consider integrating at least part of this patch into master.

PR #270, opened 2017-02-03
Switch default docker exec command back to `bash {script}`

meatballhat:
but allow for setting it via env for testing.

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -26,10 +26,12 @@ import (
 var (
 	defaultDockerNumCPUer       dockerNumCPUer = &stdlibNumCPUer{}
 	defaultDockerSSHDialTimeout                = 5 * time.Second
+	defaultExecCmd                             = "bash /home/travis/build.sh"
 	dockerHelp                                 = map[string]string{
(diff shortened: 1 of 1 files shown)


igorwwwwwwwwwwwwwwwwwwww:
perhaps make it clear in the help string that it only applies to docker native mode?

backend/docker.go
```diff
@@ -506,7 +506,7 @@ func (i *dockerInstance) runScriptSSH(ctx gocontext.Context, output io.Writer) (
 	}
 	defer conn.Close()
 
-	exitStatus, err := conn.RunCommand("bash ~/build.sh", output)
+	exitStatus, err := conn.RunCommand(strings.Join(i.provider.execCmd, " "), output)
```
igorwwwwwwwwwwwwwwwwwwww: 🤦‍♀️

igorwwwwwwwwwwwwwwwwwwww: oh you're adding it to the other one too!

igorwwwwwwwwwwwwwwwwwwww: 👍

PR #271, opened 2017-02-03
replace golang.org/x/net/context with the native go 1.7 context

igorwwwwwwwwwwwwwwwwwwww:


processor.go
```diff
@@ -71,6 +71,7 @@ func NewProcessor(ctx gocontext.Context, hostname string, queue JobQueue,
 	buildJobsChan, err := queue.Jobs(ctx)
 	if err != nil {
 		context.LoggerFromContext(ctx).WithField("err", err).Error("couldn't create jobs channel")
+		cancel()
```
igorwwwwwwwwwwwwwwwwwwww: please pay attention to this change and let me know if that looks fine to you

meatballhat: I'm assuming this means it was missing and you noticed, but that it's not strictly related to the switch to stdlib `context`, ja? 😄

igorwwwwwwwwwwwwwwwwwwww: govet started complaining about it, so I guess they have a check that only works with stdlib. :)

PR #273, opened 2017-02-03
job: log requeue event and finish state

igorwwwwwwwwwwwwwwwwwwww:
We've had trouble figuring out from logs what the finish state was in the past, I figured it'd be useful to log it explicitly.


http_job.go
```diff
@@ -80,10 +81,12 @@ func (j *httpJob) Error(ctx gocontext.Context, errMessage string) error {
 		return err
 	}
 
-	return j.Finish(FinishStateErrored)
+	return j.Finish(ctx, FinishStateErrored)
 }
 
-func (j *httpJob) Requeue() error {
+func (j *httpJob) Requeue(ctx gocontext.Context) error {
+	context.LoggerFromContext(ctx).WithField("job", j.Payload().Job.ID).Info("requeueing job")
```
henrikhodne: The `job` field should be set by `LoggerFromContext`, or is this context coming from somewhere else?

igorwwwwwwwwwwwwwwwwwwww: I don't think we are currently setting it on the context. I wasn't able to find any reference to `context.WithValue()`. I agree that we should do that though, that would clean up a whole lot of code.

henrikhodne: https://github.com/travis-ci/worker/blob/31dbcbed3e4da64504c087dcff1c05bc8b6eadd9/context/package.go#L53-L55 and https://github.com/travis-ci/worker/blob/31dbcbed3e4da64504c087dcff1c05bc8b6eadd9/processor.go#L141 sets it

igorwwwwwwwwwwwwwwwwwwww: Ah, github's search compacted the results. Thanks!

igorwwwwwwwwwwwwwwwwwwww:
@henrikhodne addressed!

PR #274, opened 2017-02-06
backend/gce: introduce PUBLIC_IP_CONNECT flag so that we can connect to the private ip from worker

igorwwwwwwwwwwwwwwwwwwww:
Currently we have a single flag `PUBLIC_IP` which controls two things:

* whether build instances receive a public IP
* whether we connect to that public ip or use the private one

It turns out however, that will be wanting to run in a hybrid mode, where instances have a public ip (so that it can talk to the internet without NAT), but no incoming traffic is allowed (dropped via firewall).

So we want to have a public IP, but not connect to it. This PR adds that capability.

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -78,6 +78,8 @@ var (
 		"PREEMPTIBLE":           "boot job instances with preemptible flag enabled (default true)",
 		"PREMIUM_MACHINE_TYPE":  fmt.Sprintf("premium machine type (default %q)", defaultGCEPremiumMachineType),
 		"PROJECT_ID":            "[REQUIRED] GCE project id",
+		"PUBLIC_IP":             "boot job instances with a public ip, disable this for NAT (default true)",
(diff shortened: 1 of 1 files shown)


backend/gce.go
```diff
@@ -930,23 +939,23 @@ func (i *gceInstance) sshConnection(ctx gocontext.Context) (ssh.Connection, erro
 }
 
 func (i *gceInstance) getIP() string {
-	// if instance has no public IP, return first private one
-	if !i.ic.PublicIP {
+	if i.ic.PublicIP && i.ic.PublicIPConnect {
```
meatballhat: Is the `PublicIPConnect` bit even needed?  This looks to me like `PublicIP` is enough.

igorwwwwwwwwwwwwwwwwwwww: @meatballhat ah, you can't see it in this diff. `PublicIP` is used in one more location in this file, when creating the instance.

PR #278, opened 2017-02-09
Linter fixes

henrikhodne:
This includes some minor-ish style fixes and dead code removal as reported by various linters in `gometalinter`. See the commit messages for more information on the various steps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,9 +33,7 @@ export DOCKER_DEST
 COVERPROFILES := \
 	backend-coverage.coverprofile \
 	config-coverage.coverprofile \
-	context-coverage.coverprofile \
-	image-coverage.coverprofile \
-	metrics-coverage.coverprofile
+	image-coverage.coverprofile
 CROSSBUILD_BINARIES := \
 	build/darwin/amd64/travis-worker \
 	build/linux/amd64/travis-worker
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -33,9 +33,7 @@ export DOCKER_DEST
 COVERPROFILES := \
 	backend-coverage.coverprofile \
 	config-coverage.coverprofile \
-	context-coverage.coverprofile \
-	image-coverage.coverprofile \
-	metrics-coverage.coverprofile
+	image-coverage.coverprofile
```
meatballhat: Is this removal because `context` and `metrics` stub tests have been removed?

http_log_writer.go
```diff
@@ -102,21 +102,6 @@ func (w *httpLogWriter) Close() error {
...
-
-	close(w.closeChan)
-	w.flush()
-
-	part = httpLogPart{
-		JobID:  w.jobID,
-		Number: w.logPartNumber,
-		Final:  true,
-	}
-	w.logPartNumber++
```
meatballhat: Huh, so this was in there twice for funsies? 😆

Makefile
henrikhodne: Correct, the tests were the stub tests and after removing them, there were no more tests (so `make test` broke since it couldn't find those files).

PR #279, opened 2017-02-09
metrics: reduce sample size of timer ExpDecaySample to have less skewed timer metrics

igorwwwwwwwwwwwwwwwwwwww:
The timer metrics we get tend to be higher than expected. I believe this is because go-metrics samples percentiles from 1024 data points, yet the job throughput is quite low -- and so high values stick around longer, biasing the percentiles upwards.


metrics/package.go
```diff
@@ -14,15 +14,29 @@ func Mark(name string) {
 
 // TimeSince increases the timer metric with the given name by the time since the given time
 func TimeSince(name string, since time.Time) {
+	metrics.DefaultRegistry.Register(name, newTimerWithSmallerSampleSize)
```
meatballhat: Is this right for the `.Register` to happen every time these funcs are called?  I would have expected a package-level registration instead, such as in an `init()` func.

henrikhodne: I think `Register` will cause an overwrite. Should probably be `RegisterOrGet` instead?

igorwwwwwwwwwwwwwwwwwwww: Nope, I checked it specifically (though I will admit it is a bit implicit behaviour). go-metrics returns a `DuplicateMetric` if the name already exists: https://github.com/rcrowley/go-metrics/blob/master/registry.go#L130

henrikhodne:
I don't think 50 samples is enough to get any accuracy with the percentiles we report, since that means we effectively only get the 0th, 2nd, 4th, etc. up to 96th, 98th, 100th percentiles, but we report the 50th, 75th, 90th, 95th, 99th, 99.9th and 100th percentiles. To get enough samples where the 99th, 99.9th and 100th percentiles all get different values we'd need at least 1000 samples.

Another option is to reduce the number of percentiles we report.

igorwwwwwwwwwwwwwwwwwwww:
After looking into the implementation a bit more, there is a second parameter (apart from the sampling reservoir size) that we can tweak: a parameter `alpha` that is a factor for how quickly values decay. It defaults to `0.015`, and from what I can tell the calculation used for the decay is roughly:

```
exp(seconds*alpha) / rand.Float64(),
```

I am not quite sure how the random value plays into this, but adjusting the alpha value might allow us to decay "old" values more quickly.

igorwwwwwwwwwwwwwwwwwwww:
Side note: build is currently failing due to https://github.com/travis-ci/travis-ci/issues/7332.

igorwwwwwwwwwwwwwwwwwwww:
This PR to go-metrics shows that the "re-scale threshold" is hard-coded to 1 hour, which might also be one of the reasons for the results we are getting: https://github.com/rcrowley/go-metrics/pull/72

igorwwwwwwwwwwwwwwwwwwww:
The more I've looked into this, the less I think what go-metrics provides gives us are meaningful numbers. At this point I would be inclined to look into using [the prometheus client](https://github.com/prometheus/client_golang), and then exporting the metrics it collects to librato.

Percentiles would work slightly differently there -- via pre-defined buckets. So some re-thinking necessary, but IMO worth investigating.

meatballhat:
Please holler if you're interested in picking this back up 🙌

PR #282, opened 2017-02-13
Speed up Travis CI builds & re-fetch deps when manifest updates

henrikhodne:
Disables tests without race detection, and changes the Makefile a bit to make sure that `gvt rebuild` is re-run if the manifest has updated since  the last time it was run.


igorwwwwwwwwwwwwwwwwwwww:
Awesome! 💜 👍

PR #290, opened 2017-03-14
picks up fixes of golang/go#18861

shawnzhu:
## Problem

We've seen lots of errors from Travis worker like this:

```
textPayload: "time="2017-03-14T19:10:57Z" level=error msg="couldn't run script, attempting requeue" completed=false err="couldn't connect to SSH server: couldn't connect to SSH server: ssh: handshake failed: ssh: unexpected message type 3 (expected one of [6])"
```

the build will finally work after requeue build request several times.

I found a fix from https://github.com/golang/go/issues/18861 and that fix stays on master since Feb 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ addons:
 install:
 - go get github.com/FiloSottile/gvt
 - go get -u github.com/alecthomas/gometalinter
+- go get -u golang.org/x/crypto/ssh
 - gometalinter --install
 - gem install package_cloud --no-ri --no-rdoc
 


igorwwwwwwwwwwwwwwwwwwww:
Nice find!

Our dependencies are managed via [gvt](https://github.com/FiloSottile/gvt), and vendored. So running `gvt update golang.org/x/crypto/ssh` to update `vendor/manifest` should fix this in a way that works with how we build worker.

henrikhodne:
Thank you for this! Like @igorwwwwwwwwwwwwwwwwwwww mentioned, we're using gvt to manage our dependencies, so I opened a new PR that updates it there instead at #292. Closing this in favor of that PR.

PR #291, opened 2017-03-15
amqp-job: send queued_at in state updates if known

### LINKED ISSUE #289: State update messages do not include the queued_at timestamp
We've agreed it would be good to include all known timestamps to all state update messages sent to Hub, but it seems we've missed the `job.queued_at` timestamp at that point.

For example, this is a `finished` message received in Hub on staging:

```
{:finished_at=>"2017-03-14T18:11:04Z", :id=>548890, :received_at=>"2017-03-14T18:10:46Z", :started_at=>"2017-03-14T18:10:47Z", :state=>"passed"}
```

It seems Scheduler includes this timestamp to its own message sent to Worker: https://github.com/travis-ci/travis-scheduler/blob/master/spec/travis/scheduler/serialize/worker_spec.rb#L64

Would it be possible for this timestamp to be included in state update messages sent back to Hub?

henrikhodne:
Fixes #289.

@svenfuchs As far as I can tell, we send no timestamps on the `reset` event, is that correct?


henrikhodne:
> How hard would it be to also add the queued_at and received_at timestamps to the reset message? They're both known at that time, aren't they?

Adding them isn't hard, I just noticed that they weren't included at all right now so I wasn't sure if that should change or not.

henrikhodne:
That actually made it easy to clean up the logic quite a bit, since the logic to make the payload is now the same for every event, only the state and "event" differs.

PR #294, opened 2017-03-21
Fix panic occuring when AMQP delivery channel is closed

henrikhodne:
When the channel is closed, `delivery` would have the zero value, which would fail JSON parsing (since the body is empty), and then panic on `Nack`.

### DIFF
--- a/amqp_job_queue.go
+++ b/amqp_job_queue.go
@@ -69,6 +69,7 @@ func (q *AMQPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err erro
 
 	go func() {
 		defer channel.Close()
+		defer close(buildJobChan)
 
 		for {
 			if ctx.Err() != nil {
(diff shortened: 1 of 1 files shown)


amqp_job_queue.go
```diff
@@ -69,6 +69,7 @@ func (q *AMQPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err erro
 
 	go func() {
 		defer channel.Close()
+		defer close(buildJobChan)
```
henrikhodne: This is where the Processor checks for channel closing: https://github.com/travis-ci/worker/blob/497c38318f6aebc5b67726b17298dd50ec56de55/processor.go#L129-L133.

Also, this is guaranteed not to panic since we only ever return a "receive-only" version of this channel from the `Jobs` method, so it's not possible to close outside of this method, and this is the only place where the channel is closed inside this method.

PR #296, opened 2017-04-02
/run mounted in container as exec

grooverdan:
Attempts to fix travis-ci/travis-ci/issues/7062 to enable build systems that generate executable contents and data on shm to be able to execute it.

The go-dockerclient supports this https://github.com/fsouza/go-dockerclient/commit/be34fed0d2cfd8c02f8ecb2bebbc7ce1cd3b9c03

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -236,7 +236,7 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 		Privileged: p.runPrivileged,
 		Memory:     int64(p.runMemory),
 		ShmSize:    int64(p.runShm),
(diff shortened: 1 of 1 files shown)


meatballhat:
@grooverdan Ahoy!  Thanks for the PR 😺  I like this, but I want to make sure it's configurable at runtime.  Do you want to take a crack at doing that, or are you OK with me basing work on top of your change?

grooverdan:
Please just base your work on top of this.

I really don't have a test environment and was flying blind a bit. Also also don't know how configurable you want it or where all the code bits that make it configurable are. Based on that, letting you take over is the sanest approach.

PR #297, opened 2017-04-05
Add job_id to build script request query string for correlation

meatballhat:


build_script_generator.go
```diff
@@ -121,6 +121,13 @@ func (g *webBuildScriptGenerator) Generate(ctx gocontext.Context, job Job) ([]by
 		u.User = nil
 	}
 
+	jp := job.Payload()
+	if jp != nil {
+		q := u.Query()
+		q.Add("job_id", fmt.Sprintf("%d", jp.Job.ID))
```
henrikhodne: Should this be `Set` instead of `Add`? It's only going to make a difference if the URL already has `job_id` for some reason, which I think is unlikely, but if it is set we'd effectively be sending an array instead of a single value.

henrikhodne: Also, `strconv.Itoa(jp.Job.ID)` would avoid the `fmt` format string parsing part. I don't think it's needed for performance reason, but it might make the readability a bit better?

PR #299, opened 2017-04-10
Merging from grooverdan/worker patch-1

meatballhat:
Closes #296


henrikhodne:
I'm not too familiar with how these Docker mounts work. Is this mounting something from the host into the container, or is it creating some kind of new tmpfs per container and mounting that?

meatballhat:
@henrikhodne My understanding is that this is tmpfs being created on demand and mounted by the docker daemon (unverified).

henrikhodne:
That looks right: https://github.com/docker/docker/pull/13587

grooverdan:
thanks @meatballhat for finishing this off and @henrikhodne for your review.

PR #307, opened 2017-05-15
Ensure AMQP job "finished_at" is included when available

meatballhat:
This addresses a regression found when upgrading from v2.5.x.

### DIFF
--- a/amqp_job_test.go
+++ b/amqp_job_test.go
@@ -182,7 +182,26 @@ func TestAMQPJob_createStateUpdateBody(t *testing.T) {
 
 	assert.Equal(t, "foo", body["state"])
 
-	for _, key := range []string{"id", "state", "meta", "queued_at", "received_at", "started_at", "finished_at"} {
+	for _, key := range []string{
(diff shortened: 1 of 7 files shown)


amqp_job_test.go
```diff
@@ -196,7 +197,7 @@ func TestAMQPJob_createStateUpdateBody(t *testing.T) {
 	job.received = time.Time{}
 	assert.NotContains(t, job.createStateUpdateBody("foo"), "received_at")
 
-	job.Payload().Job.QueuedAt = &time.Time{}
+	job.Payload().Job.QueuedAt = nil
```
Lyoness: Hi @meatballhat I understood everything else, but this. 
Why `nil` ?  Because we're moving up the `QueuedAt` and assigning it to time.Time up at #L53?

meatballhat: The `QueuedAt` field is deserialized from JSON, so I'm pretty sure we use `*time.Time` (instead of `time.Time`) as a defensive measure.

PR #311, opened 2017-05-23
HTTP job queue debugging and visibility improvements

meatballhat:


http_job_queue.go
```diff
@@ -58,34 +59,54 @@ func NewHTTPJobQueue(pool *ProcessorPool, jobBoardURL *url.URL, site, providerNa
 
 // Jobs consumes new jobs from job-board
 func (q *HTTPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err error) {
+	logger := context.LoggerFromContext(ctx)
+	if q.buildJobChan != nil {
+		return q.buildJobChan, nil
+	}
```
igorwwwwwwwwwwwwwwwwwwww: Is .Jobs() only called from one goroutine? If not, this looks like it might be a potential race condition.

http_log_writer.go
```diff
@@ -57,7 +57,11 @@ func newHTTPLogWriter(ctx gocontext.Context, url string, authToken string, jobID
 		httpClient: &http.Client{},
 		baseURL:    url,
 		authToken:  authToken,
-	}, nil
+	}
+
+	go writer.flushRegularly(ctx)
```
igorwwwwwwwwwwwwwwwwwwww: 👍

http_job_queue.go
meatballhat: I thought it was all in serial, but I was wrong!  Adding a mutex....

PR #312, opened 2017-05-24
Add multiple source job queue type

meatballhat:
and handle comma-delimited queue type values as multiple source queues.

The goal here is to be able to run worker instances in production with a config like the following so that we can gradually introduce more HTTP-based traffic without having to separately manage HTTP-based capacity in all infrastructures:

``` bash
export TRAVIS_WORKER_QUEUE_TYPE=amqp,http
```

- [x] more test coverage dangit
- [x] does it work?

### DIFF
--- a/multi_source_job_queue.go
+++ b/multi_source_job_queue.go
@@ -1,11 +1,13 @@
 package worker
 
 import (
+	"strings"
 	"sync"
 	"time"
 
 	gocontext "context"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/travis-ci/worker/context"
 )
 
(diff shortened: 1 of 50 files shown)


multi_source_job_queue.go
```diff
@@ -0,0 +1,65 @@
...
+type MultiSourceJobQueue struct {
+	queues             []JobQueue
+	buildJobChans      []<-chan Job
+	buildJobChansMutex *sync.Mutex
+}
+
+func NewMultiSourceJobQueue(queues ...JobQueue) *MultiSourceJobQueue {
+	return &MultiSourceJobQueue{
+		queues:             queues,
+		buildJobChansMutex: &sync.Mutex{},
```
igorwwwwwwwwwwwwwwwwwwww: This is unused

multi_source_job_queue.go
```diff
@@ -0,0 +1,65 @@
...
+	buildJobChan := make(chan Job)
+	outChan = buildJobChan
+
+	go func() {
+		tjq.buildJobChans = []<-chan Job{}
+		for _, queue := range tjq.queues {
+			jc, err := queue.Jobs(ctx)
+			if err != nil {
+				logger.WithField("err", err).Error("failed to get job chan from queue")
+				return
```
igorwwwwwwwwwwwwwwwwwwww: I believe this logic allows any failure in any `queue.Jobs()` call to break the entire `Jobs()` from succeeding. I might be missing a part where this is handled though.

meatballhat: That's the behavior that I intend, yep!  Any queue having an issue is cause for teardown.

multi_source_job_queue.go
meatballhat: Tragic!  fixing...

meatballhat:
On a worker with `TRAVIS_WORKER_QUEUE_TYPE=http,amqp`:

- amqp job: https://staging.travis-ci.org/travis-repos/ruby-test-staging/jobs/592785
- http job: https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/jobs/592790

meatballhat:
Even though logs appear to be flowing correctly in staging, the log part sink buffer is filling up 😕, e.g.:

```
time="2017-06-06T14:08:21Z" level=error msg="could not add log part to sink" component="log_writer" err="log sink buffer reached maximum size 150" job_id=592793 pid=1 processor=0467faf9-48ce-48e9-92d0-39d4383b2278 repository="meatballhat/yolo-octo-adventure" self="http_log_writer"
```

meatballhat:
There's a bit of a concept mismatch going on (for me, at least) with the HTTP log part sink such that there's a tree of stuff created based on the scope of a job right down to the HTTP log writer, and _then_ we need to get a hold of an HTTP log part sink that is shared.  For awhile I was banging my head against an issue in staging where a given HTTP log part sink's buffer would stop being flushed after a few minutes, which I believe is because the `<-ctx.Done()` case was selectable as the `ctx` was, at the time, scoped to a job.  The solution present in this patch includes the addition of a `rootContext`, which is a stinky awful global variable (package-level variable, really), but seemed better-ish to me than tacking a circular reference onto the context itself as a value.  Opinions.

meatballhat:
yeap I broke something; fixin'

PR #314, opened 2017-05-25
Batch together HTTP log parts writes

meatballhat:
mostly to cut down a bit on the bulk number of HTTP requests.  I'm guessing that we may want the max batch size to be configurable (?).  **NOTE**: I couldn't find where or how the http log part `UUID` was being used, which is why I dropped it.

- [x] https://github.com/travis-ci/travis-logs/pull/131 is merged
- [x] is this a good idea?
- [x] does it work?
- [x] more tests!

**Update**: This solution explicitly _does not_ increase the size of a given log part because there is a max size restriction imposed by Pusher.  Further explanation is available [here](https://github.com/travis-ci/worker/blob/51222dac0393503fc5a28657e2997273e56d6ca1/log_writer.go#L14-L32)! :heart:

### DIFF
--- a/http_log_part_sink.go
+++ b/http_log_part_sink.go
@@ -13,6 +13,7 @@ import (
 	gocontext "context"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/cenkalti/backoff"
 	"github.com/pkg/errors"
 	"github.com/travis-ci/worker/context"
 )
@@ -35,9 +36,10 @@ const (
 )
 
 type httpLogPartSink struct {
-	httpClient *http.Client
-	baseURL    string
-	authToken  string
+	httpClient  *http.Client
+	httpBackOff *backoff.ExponentialBackOff
+	baseURL     string
+	authToken   string
 
 	partsBuffer      []*httpLogPart
 	partsBufferMutex *sync.Mutex
@@ -46,8 +48,14 @@ type httpLogPartSink struct {
 }
 
 func newHTTPLogPartSink(ctx gocontext.Context, url, authToken string, maxBufferSize uint64) *httpLogPartSink {
+	httpBackOff := backoff.NewExponentialBackOff()
+	// TODO: make this configurable?
+	httpBackOff.MaxInterval = 10 * time.Second
+	httpBackOff.MaxElapsedTime = 3 * time.Minute
+
 	lps := &httpLogPartSink{
 		httpClient:       &http.Client{},
+		httpBackOff:      httpBackOff,
 		baseURL:          url,
 		authToken:        authToken,
 		partsBuffer:      []*httpLogPart{},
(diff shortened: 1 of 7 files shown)


http_log_writer.go
```diff
@@ -191,48 +200,68 @@ func (w *httpLogWriter) flush() {
...
+	for _, part := range parts {
+		payload = append(payload, &httpLogPartPayload{
+			Type:     "log_part",
+			Final:    part.Final,
+			Content:  base64.StdEncoding.EncodeToString([]byte(part.Content)),
+			Encoding: "base64",
+		})
+	}
+
+	return w.publishLogPartOrParts(payload, "POST", "multi")
```
igorwwwwwwwwwwwwwwwwwwww: Does the travis-logs endpoint handle `/multi` requests specially? IMO this should be a separate endpoint on the travis-logs side and also a separate method on the golang side.

igorwwwwwwwwwwwwwwwwwwww:
Overall question: What benefit does this provide over making log parts larger?

http_log_writer.go
meatballhat: I'm prepping the travis-logs PR now, given last night's hotel Wi-Fi sadness 👍

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww I attempted to address your overall question in the PR description 😁

igorwwwwwwwwwwwwwwwwwwww:
@meatballhat that makes sense, thanks! Could it make sense to batch log parts _across_ processors in worker?

http_log_writer.go
```diff
@@ -191,48 +200,68 @@ func (w *httpLogWriter) flush() {
...
+func (w *httpLogWriter) logFlushError(err error) {
+	if err == nil {
+		return
 	}
 
-	payload := logPartPayload{
+	context.LoggerFromContext(w.ctx).WithFields(logrus.Fields{
+		"err":  err,
+		"self": "http_log_writer",
+	}).Error("couldn't publish log parts")
```
meatballhat: Looks like we're already covered with the logrus sentry hook if `sentry-hook-errors` is enabled. 👍

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww Batching across processors SGTM, but will def require some rework of what "owns" what, possibly introducing a shared buffer 🙀.

http_log_writer.go
igorwwwwwwwwwwwwwwwwwwww: I would double check if that is the case, because I do remember adding custom sentry handling, which implies that it may not be the case.

joshk:
will this reduce how often log parts are sent to travis-logs?

meatballhat:
@joshk If we do it wrong, maybe!  Based on feedback from @igorwwwwwwwwwwwwwwwwwwww, I think we're collectively more in favor of buffering log parts across all processors rather than per-processor (per job scope), which I would expect to help minimize job-level log stuttering.

http_log_part_sink.go
```diff
@@ -0,0 +1,145 @@
...
+		})
+	}
+
+	err := lps.publishLogParts(payload)
+	if err != nil {
+		context.LoggerFromContext(ctx).WithFields(logrus.Fields{
+			"self": "http_log_part_sink",
+			"err":  err,
+		}).Error("failed to publish buffered parts")
+		return
```
igorwwwwwwwwwwwwwwwwwwww: maybe we could add a comment here explaining that this is where the fault tolerance for log part publishing errors happens, since it's quite an important aspect of the `partsBuffer`.

meatballhat: GitHub didn't like my comment placement enough to resolve your note 🤷‍♂️

Issue #353, opened 2017-08-17
Multi-Source Job Queue holds a single job


meatballhat:
When the multi-source job queue is used, such as when the queue type is configured as a comma-delimited value, there is a single job held in a variable between input channel recv and output channel send.  This results in "stuck job" behavior whenever all available processors are busy.

PR #354, opened 2017-08-17
Send to job queue only when ready

meatballhat:
### What is the problem that this PR is trying to fix?

See #353

### What approach did you choose and why?

I decided to change the relationship between individual processors and job-board so that claims and delivery are handled at the processor level rather than being batched together through a shared job queue.  This is very similar to the way the AMQP job queue works.

### How can you test this?

I would like to deploy this to AWS staging with queue type of `amqp,http`.

### DIFF
--- a/amqp_job_queue.go
+++ b/amqp_job_queue.go
@@ -89,7 +89,6 @@ func (q *AMQPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err erro
 			case <-ctx.Done():
 				return
 			case <-time.After(time.Second):
-				logger.Debug("timeout waiting for delivery")
 				continue
 			case delivery, ok := <-deliveries:
 				if !ok {
--- a/amqp_job_queue.go
+++ b/amqp_job_queue.go
@@ -51,7 +51,7 @@ func NewAMQPJobQueue(conn *amqp.Connection, queue string) (*AMQPJobQueue, error)
 // Jobs creates a new consumer on the queue, and returns three channels. The
 // first channel gets sent every BuildJob that we receive from AMQP. The
 // stopChan is a channel that can be closed in order to stop the consumer.
-func (q *AMQPJobQueue) Jobs(ctx gocontext.Context, ready <-chan struct{}) (outChan <-chan Job, err error) {
+func (q *AMQPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err error) {
 	channel, err := q.conn.Channel()
 	if err != nil {
 		return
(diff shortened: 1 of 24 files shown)


amqp_job_queue.go
```diff
@@ -84,7 +84,8 @@ func (q *AMQPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err erro
 			select {
 			case <-ctx.Done():
 				return
-			case delivery, ok := <-deliveries:
+			case <-ready:
+				delivery, ok := <-deliveries
```
meatballhat: TODO: I think that this recv from `<-deliveries` should also have a `select` timeout.

http_job_queue.go
```diff
@@ -84,29 +84,31 @@ func (q *HTTPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err erro
...
+							}).Info("sent job to output channel")
+						}
+					case <-time.After(100 * time.Millisecond):
+						logger.Debug("timeout waiting for ready chan")
 					}
 				}
 			}
 
 			select {
-			case <-time.After(time.Second):
```
meatballhat: This is a redundant `1s` sleep, fwiw.

http_job_queue.go
```diff
@@ -84,29 +84,31 @@ func (q *HTTPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err erro
 				time.Sleep(time.Second)
 			} else {
 				for _, id := range jobIds {
```
meatballhat: This `jobIds` slice should only contain as many job IDs as there are ready processors, although in the case of a multi-source queue in front of an HTTP and AMQP queue, the ready processors count may be incorrect, which is why each ID is additionally guarded by the select over `<-ready`.

meatballhat:
⚠️ This is very much _not_ working correctly in staging, ftr 🎉

multi_source_job_queue.go
```diff
@@ -37,26 +41,40 @@ func (msjq *MultiSourceJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job
...
+				logger.Debug("about to receive job")
 				select {
-				case job := <-bjc:
-					buildJobChan <- job
-					metrics.TimeSince("travis.worker.job_queue.multi.blocking_time", jobSendBegin)
-					logger.WithFields(logrus.Fields{
-						"source": queueName,
-						"dur":    time.Since(jobSendBegin),
-					}).Info("sent job to multi source output channel")
+				case job = <-bjc:
```
bogdanap: I think we could keep job logging and sending inside this clause. It shouldn't change the behaviour and it feels easier to follow (at least for me).

processor.go
```diff
@@ -140,24 +137,39 @@ func (p *Processor) Run() {
 				return
 			}
 
+			if buildJob == nil {
```
bogdanap: Just out of curiosity, when would this happen (since we're actually crafting the buildJob object before it ends up on the buildJobsChan)?

http_job_queue.go
```diff
@@ -51,163 +53,178 @@ type jobBoardErrorResponse struct {
...
-	fetchRequestPayload := &httpFetchJobsRequest{Jobs: []string{}}
-	numWaiting := 0
-	q.processorPool.Each(func(i int, p *Processor) {
-		switch p.CurrentStatus {
-		case "processing":
-			fetchRequestPayload.Jobs = append(fetchRequestPayload.Jobs, strconv.FormatUint(p.LastJobID, 10))
-		case "waiting", "new":
-			numWaiting++
-		}
+func (q *HTTPJobQueue) pollForJob(ctx gocontext.Context, buildJobChan chan Job) bool {
```
bogdanap: I think some docstrings for this and all the other new functions would help make things clearer.

processor.go
meatballhat: You're right that this shouldn't happen.  I think it's a leftover from an earlier iteration where I was trying to do a ready channel at the processor level.

meatballhat:
This is working correctly in staging 👍

PR #356, opened 2017-08-24
Add defaultAllowReauth as an option to be set when connecting to Openstack

rpsene:
We have seen in worker logs msgs like below:
    "time="2017-08-21T13:36:00Z" level=error msg="couldn't start instance"
    *err="couldn't get image at Openstack backend: Unable to find image:"

    This error is caused because the Openstack driver interface of worker is
    initializing all the service client (keystone/nova/image) at the starting
    of the worker and these client objects are re-used during its entire
    life cycle. After a certain period of time, the token generated by the
    identity service expires and these client objects are not able to fulfill
    any request until worker is fully restarted. To overcome this issue
    "AllowReauth: true" needs to be added during Openstack client
    initialization to allow re-authentication to happen automatically.

    According the Openstack documentationi [1], if allow_reauth is True
    and connection token is going to expire soon then the method returns
    updated connection. The method invariant is the following: if
    self.allow_reauth is False then the method returns the same
    connection for every call. So the connection may expire.
    If self.allow_reauth is True the returned swift connection
    is always valid and cannot expire at least for
    swift_store_expire_soon_interval.

    [1] - https://docs.openstack.org/glance_store/latest/reference/api/\
    glance_store._drivers.swift.connection_manager.html

Signed-off-by: Spurti Chopra <spurti@us.ibm.com>

### DIFF
--- a/backend/openstack.go
+++ b/backend/openstack.go
@@ -46,6 +46,7 @@ const (
 	defaultOSSSHUser           = "travis"
 	defaultOSKeyPairName       = ""
 	defaultOSSSHKeyPath        = ""
+	defaultAllowReauth         = "true"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


rpsene:
More work is needed.

PR #357, opened 2017-08-25
Update Dockerfile

soulshake:
Please make sure you cover the following points:

## 1. What is the problem that this PR is trying to fix?

`docker build` does not work on master due to the line adding this file, which doesn't exist in the build context:

`ADD build/linux/amd64/travis-worker /usr/local/bin/travis-worker`

## 2. What approach did you choose and why?

Updated `Dockerfile` to use a multi-stage Dockerfile which is then used as an automated build on Docker Hub.

## 3. How can you test this?

By running: `docker build -t worker .`

### Important

Presumably, this repo corresponds to the [travisci/worker](https://hub.docker.com/r/travisci/worker) image on the Docker Hub. I don't know how the hub image is built, so **merging this PR might cause it to break.**

(If someone could enlighten me about this, that would be great!)

---

Partially addresses https://github.com/travis-ci/worker/issues/358


meatballhat:
As to your question in the description, the image on docker hub is built and pushed during the CI process [here](https://github.com/travis-ci/worker/blob/c76d17cfb32f927e8c7ba12617d7d0d3431e57c0/.travis.yml#L45-L47).

Dockerfile
```diff
@@ -1,12 +1,22 @@
...
 MAINTAINER Travis CI GmbH <support+travis-worker-docker-image@travis-ci.org>
 
-ADD build/linux/amd64/travis-worker /usr/local/bin/travis-worker
-ADD .docker-entrypoint.sh /docker-entrypoint.sh
+RUN go get -u github.com/FiloSottile/gvt
 
-RUN apk add --no-cache ca-certificates curl bash
+COPY . /go/src/github.com/travis-ci/worker
+WORKDIR /go/src/github.com/travis-ci/worker
+RUN go get ./...
```
meatballhat: Shouldn't this be `make deps` instead?

soulshake:
Alright, come at me @meatballhat :) What do you think of this multi-stage setup? The resulting `travis-worker` image is just 9 MB (see [soulshake/worker-upstream](https://hub.docker.com/r/soulshake/worker-upstream/tags/)).

.travis.yml
```diff
@@ -42,13 +47,14 @@ script:
...
-  fi
-- if [[ $TRAVIS_PULL_REQUEST = 'false' && $PACKAGECLOUD_TOKEN ]]; then
-    make package smoke-package;
-  fi
+
+after_success:
+  - echo $TRAVIS_PULL_REQUEST
+  - echo $TRAVIS_BRANCH
+  - sed -i s/@@BRANCH/$TRAVIS_BRANCH/ bin/send-docker-hub-trigger.sh
+  - bin/send-docker-hub-trigger.sh
```
meatballhat: I think this script should instead be `./script/trigger-docker-hub-build` and should accept the branch as a positional argument instead of altering it with `sed`.

bin/send-docker-hub-trigger.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# This script will trigger a multi-stage Docker build at the Docker hub repo referenced by $DOCKER_HUB_TRIGGER_URL.
+
+if [[ "$TRAVIS_PULL_REQUEST" == 'false' ]]; then
+  echo "Triggering Docker Hub build on branch @@BRANCH"
+  curl -H "Content-Type: application/json" --data '{"source_type": "Branch", "source_name": "@@BRANCH"}' -X POST "$DOCKER_HUB_TRIGGER_URL"
```
meatballhat: Yep right here, instead of a `@@BRANCH` string that gets replaced, this should instead be either using `$TRAVIS_BRANCH` directly (and asserting it is present), or accepting a positional argument, e.g.:

``` bash
: "${TRAVIS_BRANCH:=${1}}"
[[ "${TRAVIS_BRANCH}" ]] || {
  echo "Missing TRAVIS_BRANCH" >&2
  exit 1
}

if [[ "${TRAVIS_PULL_REQUEST}" == 'false' ]]; then
# ...
```

Dockerfile
```diff
@@ -1,12 +1,20 @@
...
 
-ADD build/linux/amd64/travis-worker /usr/local/bin/travis-worker
-ADD .docker-entrypoint.sh /docker-entrypoint.sh
+RUN go get -u github.com/FiloSottile/gvt
 
-RUN apk add --no-cache ca-certificates curl bash
+COPY . /go/src/github.com/travis-ci/worker
+WORKDIR /go/src/github.com/travis-ci/worker
+RUN make deps
+RUN make build
```
meatballhat: I'm pretty sure the binary generated here will not work inside an alpine container as-is, which is why the separate `make crossbuild` target was previously used.  I think you should be able to get the same end result by adding a directive above of `ENV CGO_ENABLED 0`.

Dockerfile
```diff
@@ -1,12 +1,20 @@
...
 
-VOLUME ["/var/tmp"]
-ENTRYPOINT ["/docker-entrypoint.sh"]
-CMD ["travis-worker"]
-STOPSIGNAL SIGINT
```
meatballhat: Can these `VOLUME`, `ENTRYPOINT`, and `STOPSIGNAL` be retained?

.travis.yml
soulshake: Moved from `bin/` to `script/` and removed `.sh` extension. (See below re: `sed`.)

bin/send-docker-hub-trigger.sh
soulshake: For some reason I wasn't able to get the Docker Hub trigger to work unless the `--data` was surrounded by single quotes, hence the torturous `sed` operation. I can take another look at this in a bit; maybe I'll have more luck with fresh eyes.

Dockerfile
soulshake: Thanks! Added.

.travis.yml
```diff
@@ -27,12 +27,17 @@ addons:
     - travis-ci/worker/$(git describe --always --dirty --tags)
     - travis-ci/worker/$TRAVIS_BRANCH
 
+before_install:
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+  - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+  - sudo apt-get update
+  - sudo apt-get -y install docker-ce
```
meatballhat: IIUC this repository and key are already present on all Trusty images.  Was upgrading docker-ce insufficient?

script/send-docker-hub-trigger
```diff
@@ -0,0 +1,15 @@
...
+
+# This script will trigger a multi-stage Docker build at the Docker hub repo referenced by $DOCKER_HUB_TRIGGER_URL.
+
+[ -z "$TRAVIS_PULL_REQUEST" ] && echo "No TRAVIS_PULL_REQUEST env var found. Aborting." && exit 1
+[ -z "$DOCKER_HUB_TRIGGER_URL" ] && echo "No DOCKER_HUB_TRIGGER_URL env var found. Aborting." && exit 1
+
+if [[ "$TRAVIS_PULL_REQUEST" == 'false' ]]; then
+  echo "Triggering Docker Hub build on branch @@BRANCH"
+  curl -H "Content-Type: application/json" \
+    --data '{"source_type": "Branch", "source_name": "@@BRANCH"}' \
```
meatballhat: This line can become the following to allow variable interpolation without resulting in the shell interpreting the json double quotes:

``` bash
    --data '{"source_type":"Branch","source_name":"'${TRAVIS_BRANCH}'"}' \
```

soulshake: Omg, thanks

PR #365, opened 2017-09-08
Fix shellcheck warnings in bin/travis-worker-install

soulshake:

`make test` results in a few test failures due to unquoted variables.

I quoted the variables 

## How can you test this?

`make test`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
   - CHECKOUT_ROOT="$HOME/gopath/src/github.com/travis-ci/worker"
   - GO15VENDOREXPERIMENT='1'
   - REDIS_URL="redis://"
+  - SHFMT_URL="mvdan.cc/sh/cmd/shfmt"
 
 addons:
   artifacts:
@@ -26,6 +27,9 @@ addons:
     - travis-ci/worker/$TRAVIS_BUILD_NUMBER/$TRAVIS_JOB_NUMBER
     - travis-ci/worker/$(git describe --always --dirty --tags)
     - travis-ci/worker/$TRAVIS_BRANCH
+
+before_install:
+- go get -u "${SHFMT_URL}" ;
 
 install:
 - go get github.com/FiloSottile/gvt
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ env:
   - CHECKOUT_ROOT="$HOME/gopath/src/github.com/travis-ci/worker"
   - GO15VENDOREXPERIMENT='1'
   - REDIS_URL="redis://"
-  - SHFMT_URL="mvdan.cc/sh/cmd/shfmt"
 
 addons:
   artifacts:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -27,6 +28,9 @@ addons:
     - travis-ci/worker/$(git describe --always --dirty --tags)
     - travis-ci/worker/$TRAVIS_BRANCH
 
+before_install:
+- go get -u "${SHFMT_URL}" ;
```
meatballhat: I'm 👎 to asserting bleeding edge `shfmt` via CI.  I would rather pin to a released version.  I see that the tests are failing because this installation of bleeding edge `shfmt` is superseded by another installation of `v1.3.1` below.  I would like to propose that both of these be removed and instead install the binary that's provided by the project a la:

``` yaml
- curl -o ~/bin/shfmt -sSL https://github.com/mvdan/sh/releases/download/v2.0.0/shfmt_v2.0.0_linux_amd64                                                                                       
- chmod +x ~/bin/shfmt                                                                                                                                                                         
- shfmt -version
```

Incidentally (!) this is what I'm doing in another branch of this repo :grin: https://github.com/travis-ci/worker/blob/e8a7520e10d458b6f8857b13d13fe57a1932a023/.travis.yml#L34-L36

.travis.yml
```diff
@@ -31,9 +31,11 @@ install:
 - go get github.com/FiloSottile/gvt
 - go get -u github.com/alecthomas/gometalinter
 - gometalinter --install
+- curl -o ~/bin/shfmt -sSL https://github.com/mvdan/sh/releases/download/v2.0.0/shfmt_v2.0.0_linux_amd64
+- chmod +x ~/bin/shfmt
+- shfmt -version
 - git clone --branch v1.3.1 https://github.com/mvdan/sh.git "${GOPATH%%:*}/src/github.com/mvdan/sh"
 - go get github.com/mvdan/sh/...
```
meatballhat: These two lines that install `shfmt` again from the `v1.3.1` branch will have to be removed 👍

backend/docker_test.go
```diff
@@ -265,6 +265,7 @@ func TestNewDockerProvider_WithInvalidCPUSetSize(t *testing.T) {
 	defer dockerTestTeardown()
 
 	assert.NotNil(t, err)
+	assert.Equal(t, "strconv.ParseInt: parsing \"fafafaf\": invalid syntax", err.Error())
```
meatballhat: Is this a leftover from your other branch addressing cpusets?

soulshake: Yes indeed, my bad! I can take it out if you want, as it doesn't really belong in this PR, though I don't think it hurts either.

.travis.yml
soulshake: Verdammt!! TY

backend/docker_test.go
meatballhat: I don't think it hurts 😃

PR #366, opened 2017-09-08
Properly allocate CPU sets to containers

soulshake:

CPU sets aren't properly allocated to containers, meaning that some jobs use more than their share of CPUs.

I think this is a problem:

```
cpuSets, err := p.checkoutCPUSets()
if err != nil && cpuSets != "" {
return nil, err
}
```
because if the `checkoutCPUSets` func returns an error, the other value is always an empty string:

```
func (p *dockerProvider) checkoutCPUSets() (string, error) {
[...]
if len(cpuSets) != p.runCPUs {
return "", fmt.Errorf("not enough free CPUsets")
}
```
so `if err != nil && cpuSets != ""` will never be true, so the error is never caught (and in that case, containers are allocated the same CPU set as the host).

## How to test

Create some sample JSON payloads and put them in `10-created.d/`:

```
$ pwd
/home/aj/git/travis/worker/builds.ec2

$ ls -la
total 60
drwxr-xr-x  7 aj aj 4096 Sep  9 15:52 .
drwxr-xr-x 19 aj aj 4096 Sep  9 16:33 ..
drwxr-xr-x  2 aj aj 4096 Sep  9 16:15 10-created.d
drwxr-xr-x  2 aj aj 4096 Sep  9 16:15 30-received.d
drwxr-xr-x  2 aj aj 4096 Sep  9 16:16 50-started.d
drwxr-xr-x  2 aj aj 4096 Sep  9 16:16 70-finished.d
drwxr-xr-x  2 aj aj 4096 Sep  9 16:13 log
-rw-r--r--  1 aj aj 1415 Sep  9 13:59 sleep-10.json
-rw-r--r--  1 aj aj 1414 Sep  9 13:59 sleep-15.json
-rw-r--r--  1 aj aj 1415 Sep  9 13:59 sleep-20.json
-rw-r--r--  1 aj aj 1415 Sep  9 13:59 sleep-25.json
-rw-r--r--  1 aj aj 1415 Sep  9 13:59 sleep-30.json
-rw-r--r--  1 aj aj 1414 Sep  9 13:59 sleep-5.json
-rw-r--r--  1 aj aj 1415 Sep  9 15:52 sleep-50.json

$ cat sleep-10.json 
{"type":"test","vm_type":"default","queue":"builds.ec2","config":{"language":"bash","env":["SLEEP=10"],"script":"./sleep.sh $SLEEP",".result":"configured","group":"stable","dist":"trusty","os":"linux"},"env_vars":[],"job":{"id":623773,"number":"1.4","commit":"1fb2f95804ee662297c908124c74e966edc3f2c9","commit_range":"191f87a69f84...1fb2f95804ee","commit_message":"yay staging","branch":"trigger","ref":null,"tag":null,"pull_request":false,"state":"passed","secure_env_enabled":true,"secure_env_removed":false,"debug_options":{},"queued_at":"2017-09-08T12:12:36Z","allow_failure":false},"source":{"id":623769,"number":"1","event_type":"push"},"repository":{"id":60659,"github_id":102856467,"slug":"soulshake/autoscale-tester","source_url":"https://github.com/soulshake/autoscale-tester.git","api_url":"https://api.github.com/repos/soulshake/autoscale-tester","last_build_id":623769,"last_build_number":"1","last_build_started_at":"2017-09-08T12:12:37Z","last_build_finished_at":null,"last_build_duration":null,"last_build_state":"started","default_branch":"master","description":"Verify autoscaling behavior"},"ssh_key":null,"timeouts":{"hard_limit":null,"log_silence":null},"cache_settings":{"fetch_timeout":1200,"push_timeout":7200,"s3":{"access_key_id":"BLAHBLAHBLAH","aws_signature_version":"4","bucket":"travis-cache-staging-org","hostname":"s3.amazonaws.com","secret_access_key":"BLAHBLAHBLAH"},"type":"s3"}}

$ cp sleep-* 10-created.d/
```
Then run worker:
```
$ travis-worker --debug 9
```

And verify that CPUs are allocated as expected:
```
$ for cid in $(docker ps -q); do docker exec -ti $cid cat /sys/fs/cgroup/cpuset/cpuset.cpus; done
6-7
4-5
0-1
2-3
```
^ Each container should have 2 unique CPUs allocated to it. The remaining jobs should not be run until these have finished.

---

This _might_ address https://github.com/travis-pro/team-blue/issues/733, at least partially.

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -337,7 +337,7 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 		logger.WithField("err", err).Error("couldn't checkout CPUSets")
 		return nil, err
 	}
-	logger.Infof("Checked out CPU sets: %#v", cpuSets)
+	logger.WithField("cpu_sets", cpuSets).Info("checked out")
 
 	if cpuSets != "" {
 		dockerConfig.CPUSet = cpuSets
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -294,16 +294,6 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 
 	logger := context.LoggerFromContext(ctx).WithField("self", "backend/docker_provider")
 
-	cpuSets, err := p.checkoutCPUSets()
-	if err != nil {
-		logger.WithField("err", err).Error("couldn't checkout CPUSets")
-		return nil, err
-	}
-
-	logger.WithFields(logrus.Fields{
-		"cpuSets": fmt.Sprintf("%#v", cpuSets),
-	}).Debug("starting container; allocating cpuSets")
-
 	if startAttributes.ImageName != "" {
 		imageName = startAttributes.ImageName
 	} else {
@@ -339,7 +329,15 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 		Memory:     int64(p.runMemory),
 		ShmSize:    int64(p.runShm),
 		Tmpfs:      p.tmpFs,
+		CPUSet:     strconv.Itoa(p.runCPUs),
 	}
+
+	cpuSets, err := p.checkoutCPUSets()
+	if err != nil {
+		logger.WithField("err", err).Error("couldn't checkout CPUSets")
+		return nil, err
+	}
+	logger.Infof("Checked out CPU sets: %#v", cpuSets)
 
 	if cpuSets != "" {
 		dockerConfig.CPUSet = cpuSets
@@ -351,10 +349,13 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 		"host_config": fmt.Sprintf("%#v", dockerHostConfig),
 	}).Debug("creating container")
 
+	// FIXME: This doesn't seem to create the container with the Config and HostConfig
 	container, err := p.client.CreateContainer(docker.CreateContainerOptions{
 		Config:     dockerConfig,
 		HostConfig: dockerHostConfig,
 	})
+	container.Config = dockerConfig
+	container.HostConfig = dockerHostConfig
 
 	if err != nil {
 		logger.WithField("err", err).Error("couldn't create container")
@@ -385,6 +386,8 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 	go func(id string) {
 		for {
 			container, err := p.client.InspectContainer(id)
+			container.Config = dockerConfig
+			container.HostConfig = dockerHostConfig
 			if err != nil {
 				errChan <- err
 				return
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -295,9 +295,14 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 	logger := context.LoggerFromContext(ctx).WithField("self", "backend/docker_provider")
 
 	cpuSets, err := p.checkoutCPUSets()
-	if err != nil && cpuSets != "" {
+	if err != nil {
+		logger.WithField("err", err).Error("couldn't checkout CPUSets")
 		return nil, err
 	}
+
+	logger.WithFields(logrus.Fields{
+		"cpuSets": fmt.Sprintf("%#v", cpuSets),
+	}).Debug("starting container; allocating cpuSets")
 
 	if startAttributes.ImageName != "" {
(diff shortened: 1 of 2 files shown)


meatballhat:
I forgot to mention it before, but it'd be rad if you could add an entry to the `Unreleased` section of CHANGELOG.md under `Fixed`, prob with a prefix of `backend/docker:`.

soulshake:
Pushed some updates.
It seems like the config also wasn't getting passed along via `p.client.CreateContainer`. Also, when the jobs get requeued (via `docker inspect`), the cpuset doesn't seem to show up in the looked-up container.
Thoughts @meatballhat ?

bogdanap:
This is an awesome find AJ! I just wanted to add that it would be cool if these changes had some unit tests associated so we make sure we don't regress sometime in the future.

soulshake:
@bogdanap I could not agree more! I looked at the tests a bit before making this change and couldn't figure out how to add one for this. I can try again now that I'm more familiar with how the code works. Suggestions welcome!

On a related note, I'm not understanding clearly our usage of some of the following terms:

* instance/VM, as in e.g. [step_start_instance.go](https://github.com/travis-ci/worker/blob/master/step_start_instance.go): what exactly are we doing when we start/stop an "instance"? For example this line:
`logger.WithField("instance", instance).Error("skipping shutdown, VM will be left running")`
Is this referring to a container instance?
* `provider`: is this supposed to correspond to a docker daemon?
* `dockerConfig` vs `dockerHostConfig` ([`docker.Config`](https://github.com/fsouza/go-dockerclient/blob/master/container.go#L286-L288) vs [`docker.HostConfig`](https://github.com/fsouza/go-dockerclient/blob/master/container.go#L717-719)): `HostConfig` seems like it should be conceptually similar to `Config` but should apply to _all_ containers launched by the same provider, except the two structs have different fields, so I'm still not fully understanding how they should be used.

backend/docker.go
```diff
@@ -343,14 +347,19 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 	logger.WithFields(logrus.Fields{
 		"config":      fmt.Sprintf("%#v", dockerConfig),
 		"host_config": fmt.Sprintf("%#v", dockerHostConfig),
-	}).Debug("starting container")
+	}).Debug("creating container")
 
+	// FIXME: This doesn't seem to create the container with the Config and HostConfig
```
meatballhat: 🙀 wat

backend/docker.go
```diff
@@ -343,14 +347,19 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
...
 	container, err := p.client.CreateContainer(docker.CreateContainerOptions{
 		Config:     dockerConfig,
 		HostConfig: dockerHostConfig,
 	})
+	container.Config = dockerConfig
+	container.HostConfig = dockerHostConfig
```
meatballhat: Why do we have to do this and the override below?

soulshake: I don't know! But I found that the containers would actually get the CPU sets they were supposed to only when these two lines were present. (I saw similar behavior when looking up requeued jobs [here](https://github.com/travis-ci/worker/blob/e18395ce242e48a54ca7549d018fc3a34f7fc19a/backend/docker.go#L389-L390).)

It really makes no sense, though, so I have doubts.

meatballhat: Arrrrrrrgh.  I'm doubly argh about this because of my current effort to switch docker client libraries.  I'll deal with it when the time comes...

backend/docker.go
```diff
@@ -333,7 +329,15 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
...
 		Tmpfs:      p.tmpFs,
+		CPUSet:     strconv.Itoa(p.runCPUs),
+	}
+
+	cpuSets, err := p.checkoutCPUSets()
+	if err != nil {
+		logger.WithField("err", err).Error("couldn't checkout CPUSets")
+		return nil, err
 	}
+	logger.Infof("Checked out CPU sets: %#v", cpuSets)
```
meatballhat: Instead of the `{level}f` formatty methods, use fields!, e.g.:

``` go
logger.WithField("cpu_sets", cpuSets).Info("checked out")
```

soulshake:
This change has been merged [here](https://github.com/travis-infrastructure/terraform-config/pull/221) and is now being rolled out to production.

### Before

cpu-before

### After

cpu-after

Lyoness:
UH-MAYZING!  Awesome @soulshake :dancing_women:

PR #369, opened 2017-09-11
Updates from shfmt v2.0.0

meatballhat:
These are the changes present on my working copy after running `make`.  I have `shfmt` `v2.0.0` installed locally, which is the same as that installed within `.travis.yml`.  Please holler if you're getting different `shfmt` changes!


meatballhat:
This looks wrong to me.  Trying again.

PR #370, opened 2017-09-12
Add docker-build smoke-docker docker-push smoke-package back to .travis.yml

soulshake:
This change is so we can continue pushing to packagecloud.io while pushing branches to [travisci/worker-testing](https://hub.docker.com/r/travisci/worker-testing) on Docker Hub.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,12 @@ script:
 - make test
 - make crossbuild
 - make smoke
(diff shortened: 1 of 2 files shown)


meatballhat:
It looks like some of the YAML is being interpreted in an unexpected way: https://travis-ci.org/travis-ci/worker/builds/274656582#L3238-L3243

PR #362, opened 2017-09-05
Switch to official docker client library

meatballhat:
plus `shfmt` updates.

Closes #351

## What is the problem that this PR is trying to fix?

The docker client library we have been using is lacking some features such as volume mounting.  It is also generally less-maintained than the official client library.

## What approach did you choose and why?

Switch to the official client library, since it is very much maintained and contains the features we would like to integrate in the future.

## How can you test this?

The docker backend is fairly well covered with tests, and we can also deploy this in AWS staging.

- [x] does it work?

PR #373, opened 2017-09-18
    Add a new option to set binding when using Docker as backend.

rpsene:
This commit adds a new function to enable setting
    bindings between the host where Docker runs and
    the container image which is launched for building
    purposes.

    According Docker documentation: "...when you use a
    bind mount, a file or directory on the host machine
    is mounted into a container. The file or directory is
    referenced by its full or relative path on the host
    machine...[1]"

    With this approach users can set multiple bindings
    by exporting the environment variable
    TRAVIS_WORKER_DOCKER_BINDS and associating the
    bindings separated by spaces:

    export TRAVIS_WORKER_DOCKER_BINDS='/var/run/docker.sock:/var/run/docker.sock /home:/home'

    When launching travis-worker, this configuration will be mapped as
    host_config="&docker.HostConfig{Binds:[]string{"/var/run/docker.sock:/var/run/docker.sock", "/home:/home"}

    [1] - https://docs.docker.com/engine/admin/volumes/bind-mounts/

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -61,6 +61,7 @@ var (
 		"SSH_DIAL_TIMEOUT":    fmt.Sprintf("connection timeout for ssh connections (default %v)", defaultDockerSSHDialTimeout),
 		"IMAGE_SELECTOR_TYPE": fmt.Sprintf("image selector type (\"tag\" or \"api\", default %q)", defaultDockerImageSelectorType),
 		"IMAGE_SELECTOR_URL":  "URL for image selector API, used only when image selector is \"api\"",
+		"BINDS":               "Bind mount a volume (example: \"/var/run/docker.sock:/var/run/docker.sock\", default \"\")",
 	}
 )
 
@@ -85,6 +86,7 @@ type dockerProvider struct {
 
 	runPrivileged bool
 	runCmd        []string
+	runBinds      []string
 	runMemory     uint64
 	runShm        uint64
 	runCPUs       int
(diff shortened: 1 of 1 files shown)


meatballhat:
@rpsene Hello, and thank you for the PR!  I'm currently working on a PR (#362) to move to using the official docker client library, which would make this PR obsolete.  Are you willing to wait for #362 to merge before attempting to add this feature?

rpsene:
@meatballhat Yep, I can await and update this PR accordingly.

rpsene:
Hello, @meatballhat! Any progress on the PR blocking this one here?

rpsene:
Hey @meatballhat, any feedback on this PR? I already updated it base on the new Docker API.

meatballhat:
@rpsene sorry for delay!

Kazade:
Could someone tell me how to mount the build dir into the docker container? I've tried various incarnations of `export TRAVIS_WORKER_DOCKER_BINDS` before calling `docker run` but that didn't seem to work. :(

meatballhat:
@Kazade I'm sorry for not being more clear about what's going on here.  We have not added support for mounting arbitrary volumes via one's `.travis.yml`, as this would almost certainly introduce security fun times.  This feature is intended to be used by humans who are deploying `travis-worker` in their own infrastructure, such as when deploying [Travis Enterprise](https://enterprise.travis-ci.com/).

Kazade:
Right. Damn :(

Could you make it so it always mounts the TRAVIS_BUILD_DIR into the same location in the container? Would that save security issues but allow building the checkout using tools in the container?

I have this situation where I want to build a library then upload it to Google Cloud storage, but the SDK I need to build with is in a docker image.

meatballhat:
@Kazade That type of feature is generally the kind of thing we punt to the sudo-enabled infrastructure (`sudo: required`).  The container-based Linux option (via the docker backend) is meant to cover ~80% of use cases.

PR #374, opened 2017-09-18
Defer amqp channel close in the right scope

meatballhat:

Errors in the RabbitMQ server logs seemed to indicate problems with how we're ordering some AMQP operations and channel closes.  This code was recently changed, and drew our attention, so the hope is that this change will address at least one source of errors.

Move the deferred channel close into the same function scope as where it is used so that it is not closed prior to use.

## How can you test this?

Tests + running it in staging.

### DIFF
--- a/amqp_job.go
+++ b/amqp_job.go
@@ -154,7 +154,6 @@ func (j *amqpJob) sendStateUpdate(ctx gocontext.Context, event, state string) er
 	if err != nil {
 		return err
 	}
-	defer amqpChan.Close()
 
 	j.stateCount++
 	body := j.createStateUpdateBody(state)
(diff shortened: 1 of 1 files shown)


bogdanap:
Considering the issue raised here:  https://github.com/travis-pro/team-blue/issues/757, I feel like this would be a better alternative than the one we agreed upon before (https://github.com/travis-ci/worker/pull/377)

meatballhat:
@bogdanap I think that a further change may be necessary, even, to back out changes to the implementation that wasn't accounting for `<-ctx.Done()` at all, since the context that's passed into the `Finish` method will have already been canceled in the case of a hard timeout, but we still want to send a state update.

bogdanap:
@meatballhat I think it would still work since the goroutine is started before the select, so it will finish the state update, even if the function receives ctx.Done().
It might not make sense to check for cancelation in the case of a hard timeout, but it might be worth it for all the other cases, I think.

PR #376, opened 2017-09-20
Warn when shutdown signal received

soulshake:

Occasionally `travis-worker` will be in the `stop/waiting` state. This PR is intended to make it more obvious when this happens due to a signal.

For example, `SIGWINCH`--window size change--is interpreted by worker as a reason to trigger a graceful shutdown. This is nontraditional behavior and as such should be warned about.

Changed `i.logger.Info` to `i.logger.Warn` in `cli.signalHandler`.

I haven't tested this. What could possibly go wrong?

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -493,10 +493,10 @@ func (i *CLI) signalHandler() {
 		case sig := <-signalChan:
 			switch sig {
 			case syscall.SIGINT:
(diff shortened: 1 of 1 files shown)


meatballhat:
> This is nontraditional behavior and as such should be warned about.

We only used `SIGWINCH` this way because other software used it 😁.  Oh well?  Time to switch signals.

PR #377, opened 2017-09-20
Back out the amqp state update timeout cleverness

meatballhat:
as it is a suspected contributing factor to general unhappiness with AMQP jobs.

We have been seeing heightened AMQP-related errors on all infrastructures since rolling out v3.0.2.

Mostly rolling back [this change](https://github.com/travis-ci/worker/commit/778e285192fc755891afbf5d59c0776758c26db5#diff-624c34b5b2d559c68f7bb49e2653436d) to return to a state update method implementation that was working well for ~2 years.

## How can you test this?

Staging rollout tests.

### DIFF
--- a/amqp_job.go
+++ b/amqp_job.go
@@ -150,6 +150,12 @@ func (j *amqpJob) createStateUpdateBody(state string) map[string]interface{} {
 }
 
 func (j *amqpJob) sendStateUpdate(ctx gocontext.Context, event, state string) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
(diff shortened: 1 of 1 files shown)


bogdanap:
Unless the timeout itself was causing issues, wouldn't the problem also be solved by https://github.com/travis-ci/worker/pull/374?

meatballhat:
@bogdanap I'm not feeling confident enough in my understanding of the production issues to say that #374 will solve the problem.  My idea with this PR was to revert the state update method to (almost) its former implementation, while retaining a check for context cancellation.

PR #378, opened 2017-09-22
Use processor context in "Finish" state update

meatballhat:
to avoid racy issues with the job context being done.

State updates failing due to job-level context cancellation, such as happens at job timeout.  This change applies both to AMQP jobs, which we've seen appear to be running past the hard timeout, and to HTTP jobs, which we've seen running past when the `Finish` state update should be complete.

Use the processor-scope context instead so that we can still prevent resource leakage while also ensuring state updates can be sent.

## How can you test this?

Tests pass, so deploying to staging!

- [x] more tests!

### DIFF
--- a/backend/fake.go
+++ b/backend/fake.go
@@ -11,6 +11,7 @@ import (
 func init() {
 	Register("fake", "Fake", map[string]string{
 		"LOG_OUTPUT": "faked log output to write",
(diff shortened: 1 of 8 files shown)


mitchellkrogza:
Issue still persisting. Showed you some samples in https://github.com/travis-ci/worker/issues/379

PR #380, opened 2017-10-03
 Use the stack-name as image-name in containers instead of the language tag

bogdanap:
https://github.com/travis-pro/team-blue/issues/760
It is currently not easy to determine from the logs which image was used for a docker build, as opposed to GCE builds, which explicitly specify if garnet, amethyst or connie was used.

Included the image name as part of the instace information for docker builds as well. 
Since we were already using tags for this information, I decided to use the tag that included the stack-name (e.g. `travisci/ci-garnet:packer-1505167479`) instead of the language tag (e.g. `travis:ruby`) which was most frequently used.

## How can you test this?
* The updated unit-tests exhibit the correct behaviour
* Run it locally or in staging and check the instance information in the build logs

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -325,7 +325,7 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 	if startAttributes.ImageName != "" {
 		imageName = startAttributes.ImageName
 	} else {
-		imageID, err := p.imageSelector.Select(&image.Params{
+		id, err := p.imageSelector.Select(&image.Params{
 			Language: startAttributes.Language,
 			Infra:    "docker",
 		})
@@ -333,6 +333,7 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 			logger.WithField("err", err).Error("couldn't select image")
 			return nil, err
 		}
+		imageID = id
 		imageName = p.dockerImageNameForID(ctx, imageID)
 	}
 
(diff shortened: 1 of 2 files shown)


backend/docker.go
```diff
@@ -295,25 +295,22 @@ func buildDockerImageSelector(selectorType string, client *docker.Client, cfg *c
...
+		return imageID
 	}
-
-	imageID, _, err := findDockerImageByTag([]string{imageName}, images)
-	if err != nil {
-		return imageName
+	for _, image := range images {
+		if image.ID == imageID {
+			for _, tag := range image.RepoTags {
+				if strings.HasPrefix(tag, "travisci/ci-") {
```
meatballhat: Maybe make this configurable with `"travisci/ci-"` as the default?  I'm thinking of installations that use custom images...

bogdanap: @meatballhat Good point, I didn't think of that. 

How about using the RepoDigest instead (so it will be something like `travisci/ci-garnet@sha256:c759b5a71f8efec6c83e5271584e4bc8345eb6804ab0fd57a293772123edec24`) or just including all the tags in the name?

meatballhat: @bogdanap I think I prefer the `travisci/ci-{stack}:{tag}` format instead of RepoDigest 👍

bogdanap: Ok, `travisci/ci-{stack}:{tag}` it is then.
Although I looked into adding it as a config option and it feels a bit unnecessary, especially since the worker has a lot of those already. 
I am inclined to think that having the image-id as a default here is good enough for identification, but am happy to hear opinions to the contrary.

bogdanap:
New instance name displayed:
screen shot 2017-10-05 at 12 55 29

PR #382, opened 2017-10-04
Do not attempt to remove links on container delete

### LINKED ISSUE #381: Container leakage in docker backend
The docker backend on master (https://github.com/travis-ci/worker/tree/2affa3f843dee8458e34c5c02ba364400bce171f) is leaking containers due to:

> Error response from daemon: Conflict, cannot remove the default name of the container

meatballhat:
because attempting to remove them results in brokenness because we don't
create any links egad! :sweat_smile:

Also, set a container name with some meaning in it.

Closes #381 

Explicitly set `RemoveLinks: false` in the container removal config since the attempted link removal is breaking and we don't (shouldn't!) need to remove links that we never created.

## How can you test this?

- [x] staging ec2

I folded in a different container naming here because I _thought_ it was related, but in the end it seems not to be.  Is the container naming sufficiently informative?

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -21,8 +21,8 @@ import (
 	docker "github.com/docker/docker/client"
 	"github.com/docker/go-connections/tlsconfig"
 	humanize "github.com/dustin/go-humanize"
-
 	"github.com/pborman/uuid"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


meatballhat:
It works 🎉 

Container leakage plugged.  Container name + hostname changed:

meatballhat:
@bogdanap the container leakage is real 😢 (actual containers, not only names)

PR #384, opened 2017-10-06
Add a cron job to hourly check the status of docker and travis-worker

soulshake:

`docker` and `travis-worker` services sometimes get into a weird state on worker instances, and need to be `start` or `restart`ed.

This PR adds two hourly cron jobs as part of the worker install step to check the status of these services and restart them if necessary.

## How can you test this?

We'll see on staging?

Is there a better way?

Addresses https://github.com/travis-pro/team-blue/issues/704

### DIFF
--- a/bin/travis-worker-install
+++ b/bin/travis-worker-install
@@ -121,6 +121,7 @@ main() {
   __install_travis_worker
   __configure_travis_worker
   __configure_grub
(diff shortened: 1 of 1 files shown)


soulshake:
Closing in favor of https://github.com/travis-infrastructure/terraform-config/pull/251

PR #385, opened 2017-10-06
Prepare for v3.1.0 release

meatballhat:
## What is the problem that this PR is trying to fix?

Reducing the size of the Unreleased delta.

## What approach did you choose and why?

(prepare for) cutting a release!  [v3.0.2...v3.1.0](https://github.com/travis-ci/worker/compare/v3.0.2...HEAD)

## How can you test this?

- [x] staging deploy(s) + poking
- [x] production canary deploy(s)
- [x] fill in the v3.0.2...v3.1.0 changelog bits!

PR #387, opened 2017-10-10
Ensure container hostname does not exceed length limit

meatballhat:
Required for #385

Generated hostnames that exceed 64 characters result in an oci-level error inside docker, which is Not Good :tm:.

Enforce some limits to keep the generated container name and hostname at 64 chars and below.

## How can you test this?

Deploy + test on AWS staging

Is allotting 10 chars for job id sufficient?

### DIFF
--- a/backend/docker_test.go
+++ b/backend/docker_test.go
@@ -593,7 +593,7 @@ func TestDocker_containerNameFromContext(t *testing.T) {
 		},
 		{
 			r: "very-SiLlY.nAmE.wat/por-cu-pine",
-			n: fmt.Sprintf("travis-job.very-SiLlY-nAmE-wat.por-cu-pine.%v", jobID),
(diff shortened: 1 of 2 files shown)


backend/docker_test.go
```diff
@@ -593,7 +593,7 @@ func TestDocker_containerNameFromContext(t *testing.T) {
 		},
 		{
 			r: "very-SiLlY.nAmE.wat/por-cu-pine",
-			n: fmt.Sprintf("travis-job.very-SiLlY-nAmE-wat.por-cu-pine.%v", jobID),
+			n: fmt.Sprintf("travis-job.very-SiLlY-nAm.por-cu-pine.%v", jobID),
```
Lyoness: A+ :egg: :tada: :blue_heart:

PR #388, opened 2017-10-11
Add instance ID to StateUpdate metadata sent to hub

soulshake:
This PR:

* Adds instance ID to StateUpdate metadata. 
* Adds `script/publish-example-payload` to facilitate local development with AMQP.
* Updates README.md.

Addresses https://github.com/travis-pro/gcp-ddos-detector/issues/1

* Added an `instance` field to the `amqpJob` type
* If instance isn't null, copy its ID to the meta field of the body created in `createStateUpdateBody`

#### The easy way

`AMQP_URI="amqp://" go test -v . -test.run TestAMQPJob_createStateUpdateBody`

#### The hard way

* From the [aj-docker-bits](https://github.com/travis-ci/travis-hub/tree/aj-docker-bits) branch of `travis-hub`, run `docker-compose up`
* Export the environment variables that have been added to the README in this PR
* Run `travis-worker`
* Run `script/publish-example-payload` to add an example queue message
* Run `docker-compose exec rabbitmq bash`. In the container:
  * enable management plugin: `rabbitmq-plugins enable rabbitmq_management`
  * `apt update && apt install wget python`
  * `wget localhost:15672/cli/rabbitmqadmin && chmod +x /rabbitmqadmin`
* Run `./script/publish-example-payload` from your host

You should be able to see an extra key in the payload metadata by running the command below in the container:

```
root@2dffdc25e3c9:/# /rabbitmqadmin get queue=reporting.jobs.builds requeue=false -f long

--------------------------------------------------------------------------------

     routing_key: reporting.jobs.builds
        exchange: 
   message_count: 0
         payload: {"id":623773,"meta":{"instance_id":"unknown instance","state_update_count":2},"queued_at":"2017-09-08T12:12:36Z","received_at":"2017-10-11T18:23:30Z","state":"reset"}
   payload_bytes: 166
payload_encoding: string
     redelivered: False

--------------------------------------------------------------------------------
```

### On staging

`trvs redis-cli travis-staging REDIS_URL keys 'hub.instance_id_job.*'`

1. Running this locally, I get an instance ID of `null`. I assume that's because I'm not GCE. Should this work on cloud VMs?

2. If so: Do I need to do something similar for the HTTP queue as well?

### DIFF
--- a/amqp_job.go
+++ b/amqp_job.go
@@ -124,13 +124,18 @@ func (j *amqpJob) LogWriter(ctx gocontext.Context, defaultLogTimeout time.Durati
 	return newAMQPLogWriter(ctx, j.conn, j.payload.Job.ID, logTimeout)
 }
 
-func (j *amqpJob) createStateUpdateBody(state string) map[string]interface{} {
+func (j *amqpJob) createStateUpdateBody(ctx gocontext.Context, state string) map[string]interface{} {
+
 	body := map[string]interface{}{
 		"id":    j.Payload().Job.ID,
 		"state": state,
 		"meta": map[string]interface{}{
 			"state_update_count": j.stateCount,
 		},
+	}
+
+	if instanceID, ok := context.InstanceIDFromContext(ctx); ok {
+		body["meta"].(map[string]interface{})["instance_id"] = instanceID
 	}
 
 	if j.Payload().Job.QueuedAt != nil {
@@ -163,7 +168,7 @@ func (j *amqpJob) sendStateUpdate(ctx gocontext.Context, event, state string) er
 	defer amqpChan.Close()
 
 	j.stateCount++
-	body := j.createStateUpdateBody(state)
+	body := j.createStateUpdateBody(ctx, state)
 
 	bodyBytes, err := json.Marshal(body)
 	if err != nil {
--- a/amqp_job.go
+++ b/amqp_job.go
@@ -125,7 +125,6 @@ func (j *amqpJob) LogWriter(ctx gocontext.Context, defaultLogTimeout time.Durati
 }
 
 func (j *amqpJob) createStateUpdateBody(ctx gocontext.Context, state string) map[string]interface{} {
-
 	body := map[string]interface{}{
 		"id":    j.Payload().Job.ID,
 		"state": state,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
   colon-paired values in `TRAVIS_WORKER_DOCKER_BINDS`
 - http-job-queue: configurable new job polling interval
 - http-job: configurable job refresh claim interval
+- amqp-job: include instance name in state update sent to hub
 
 ### Changed
 - backend/gce: add site tag to job vms
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ## [Unreleased]
 
 ### Added
+- amqp-job: include instance name in state update sent to hub
 
 ### Changed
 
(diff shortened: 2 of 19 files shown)


example-publish.py
```diff
@@ -0,0 +1,15 @@
...
+
+connection = pika.BlockingConnection(pika.ConnectionParameters('localhost'))
+channel = connection.channel()
+
+channel.basic_publish(exchange='',
+                      routing_key=os.environ["TRAVIS_WORKER_QUEUE_NAME"]
+                      body=data)
+print(" [x] Sent 'Enqueued the thing!'")
+
+connection.close()
```
meatballhat: nitpick: I think this script should be named *without* the `.py` extension (since it is not importable), and should instead be located at `./script/publish-example-payload`, also changing the `open` to read from an absolute path, e.g.:

``` python
# ...
import os

top = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))

with open(os.path.join(top, 'example-payload.json'), 'r') as f:
    data = f.read()

# ...
```

soulshake:
I'm not really convinced this is a good approach :[ It doesn't feel right.
And in the case of sudo-enabled jobs, I get the feeling this will return the hostname of the instance _worker_ is running on, rather than the instance the _job_ is running on. No?

processor.go
```diff
@@ -14,8 +14,8 @@ import (
 // A Processor gets jobs off the job queue and coordinates running it with other
 // components.
 type Processor struct {
-	ID       string
-	hostname string
+	ID         string
+	instanceID string
```
igorwwwwwwwwwwwwwwwwwwww: I don't think we need this at the processor level anymore

processor.go
```diff
@@ -204,10 +206,11 @@ func (p *Processor) Terminate() {
 
 func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
 	state := new(multistep.BasicStateBag)
-	state.Put("hostname", p.ID)
+	state.Put("instanceID", p.ID)
 	state.Put("buildJob", buildJob)
 	state.Put("procCtx", p.ctx)
 	state.Put("ctx", ctx)
+	ctx = context.FromInstanceID(ctx, p.ID)
```
igorwwwwwwwwwwwwwwwwwwww: these can be removed, since we're using `instance.ID()` instead of `p.ID`

amqp_job.go
```diff
@@ -21,15 +21,16 @@ type amqpJob struct {
 	payload         *JobPayload
 	rawPayload      *simplejson.Json
 	startAttributes *backend.StartAttributes
+	instanceID      string
```
igorwwwwwwwwwwwwwwwwwwww: this can be removed

amqp_job.go
```diff
@@ -21,15 +21,16 @@ type amqpJob struct {
...
 	started         time.Time
 	finished        time.Time
 	stateCount      uint
 }
 
 func (j *amqpJob) GoString() string {
-	return fmt.Sprintf("&amqpJob{conn: %#v, delivery: %#v, payload: %#v, startAttributes: %#v}",
-		j.conn, j.delivery, j.payload, j.startAttributes)
+	return fmt.Sprintf("&amqpJob{conn: %#v, delivery: %#v, payload: %#v, startAttributes: %#v, instance_id: %#v}",
+		j.conn, j.delivery, j.payload, j.startAttributes, j.instanceID)
```
igorwwwwwwwwwwwwwwwwwwww: this can also be removed, as we are getting the instance id from ctx in `createStateUpdateBody`, no need to attach it to the job

processor.go
```diff
@@ -204,7 +204,7 @@ func (p *Processor) Terminate() {
 
 func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
 	state := new(multistep.BasicStateBag)
-	state.Put("hostname", p.ID)
+	state.Put("instanceID", p.ID)
```
igorwwwwwwwwwwwwwwwwwwww: I believe this is no longer needed

igorwwwwwwwwwwwwwwwwwwww:
I'd say that this is good to go! Add `CHANGELOG.md` entry and merge?

PR #389, opened 2017-10-13
add site tag to gcp provider for instance filtering

igorwwwwwwwwwwwwwwwwwwww:

We need the ability to filter instances by site in order to correctly route alerts in stackdriver to the correct webhook endpoint.

This patch allows the site to be specified via an env var, it will be attached as a network tag to all job VMs.

## How can you test this?

We can deploy it to staging and check a newly booted instance for its network tags. It should include `testing` and `site-<site>` -- e.g. `site-org`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 
 ### Changed
+- backend/gce:
+    - add site tag to job vms
 
 ### Deprecated
 
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -11,6 +11,7 @@ import (
 	mathrand "math/rand"
 	"net/http"
 	"net/url"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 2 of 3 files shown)


meatballhat:
I said something similar in the travis-infrastructure/terraform-config#255, but can you use the existing `TRAVIS_WORKER_TRAVIS_SITE` value instead of adding this?

igorwwwwwwwwwwwwwwwwwwww:
I considered passing `TRAVIS_WORKER_TRAVIS_SITE` to the provider, but decided against it, since the separation between providers and the rest of the code is quite strong and dynamically adding provider-specific variables felt messy.

I'm happy to rework it though, if you have a strong preference towards reusing the existing config value.

meatballhat:
I would like to propose adding `func (pc *ProviderConfig) Site() string` so that _all_ providers can access this value 👍

meatballhat:
On second thought, populating a value for `Site() string` that comes from `Getenv('TRAVIS_WORKER_TRAVIS_SITE')` is about as invasive (?) as populating a `SITE` key, so either one works for me.  I think that it's reasonable to provide some top-level config values in the provider config.

igorwwwwwwwwwwwwwwwwwwww:
@meatballhat I've reworked this PR to re-use the existing `TravisSite` and pass it down into the provider config.

igorwwwwwwwwwwwwwwwwwwww:
Looks like one of the tests was a little flakey. All passing now ✅ .

igorwwwwwwwwwwwwwwwwwwww:
Tested on staging, it works!

PR #394, opened 2017-10-17
Consistently construct GCE and Docker hostnames

meatballhat:

Inconsistent hostname generation between the two most-used backends in production.

Use the same hostname generation func in both cases.

## How can you test this?

The tests + canary deployment in staging.

Should this be applied to other backends where possible?


meatballhat:
👌 
screenshot_2017-10-17_18-35-09

PR #404, opened 2017-10-20
Check CPUsets back in if not needed, and be verbose about it.

soulshake:

- Container creation fails when a stray container exists with the same name.
- CPUsets aren't getting checked back in when container creation fails.

This combination results in cascading requeues (each time a job is requeued on the same instance, the worker loses track of that CPU set).

- Make worker check for a preexisting container with the same name. If it exists, delete it.
- Check CPU sets back in when container creation fails.
- Add some logging around the above.

## How can you test this?

Run some jobs. Restart docker while the jobs are running, which should result in `Exited` containers (`docker ps -a`). Logs should show CPU sets being checked back in, and `Exited` containers being deleted.

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -366,7 +366,7 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 		},
 	}
 
-	cpuSets, err := p.checkoutCPUSets()
+	cpuSets, err := p.checkoutCPUSets(ctx)
 	if err != nil {
 		logger.WithFields(logrus.Fields{
 			"err":            err,
@@ -375,7 +375,6 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 		}).Error("couldn't checkout CPUSets")
 		return nil, err
 	}
-	logger.WithField("cpu_sets", cpuSets).Info("checked out")
 
 	if cpuSets != "" {
 		dockerHostConfig.Resources.CpusetCpus = cpuSets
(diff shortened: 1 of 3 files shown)


backend/docker.go
```diff
@@ -390,6 +389,9 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
 		ctx, dockerConfig, dockerHostConfig, nil, containerName)
 
 	if err != nil {
+		logger.WithField("err", err).Error("couldn't create container")
+		defer p.checkinCPUSets(ctx, cpuSets)
```
igorwwwwwwwwwwwwwwwwwwww: what about other error paths below? e.g. `p.client.ContainerStart()` we probably want to checkin there too

igorwwwwwwwwwwwwwwwwwwww: I believe we also need it in the `ctx.Err()` case.

backend/docker.go
```diff
@@ -477,20 +496,34 @@ func (p *dockerProvider) checkoutCPUSets() (string, error) {
...
 	defer p.cpuSetsMutex.Unlock()
+	logger := context.LoggerFromContext(ctx).WithField("self", "backend/docker_provider")
 
 	for _, cpuString := range strings.Split(sets, ",") {
 		cpu, err := strconv.ParseUint(cpuString, 10, 64)
 		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"err":        err,
+				"cpu_string": cpuString,
+			}).Error("couldn't checkin CPU")
```
meatballhat: IIUC this error message should instead be `"couldn't parse CPU string"`

PR #405, opened 2017-10-24
pin docker version

soulshake:

The version of docker we're running (`17.03.1-ce`) is outdated because we only install it if it's not already present:
```
  if ! docker version &>/dev/null; then
    apt-get install -y \
      "linux-image-extra-$(uname -r)" \
      docker-ce
  fi
```

Instead of installing if not present, we install a specific version of Docker.

### DIFF
--- a/bin/travis-worker-install
+++ b/bin/travis-worker-install
@@ -185,11 +185,9 @@ __install_docker() {
 
   apt-get update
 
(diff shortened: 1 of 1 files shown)


soulshake:
Oops, this script is only used for enterprise installs. Never mind!

PR #415, opened 2017-10-31
Improve logging around requeue conditions

meatballhat:

Mildly inconsistent logging around requeue conditions, as well as requiring investigation beyond log line for relevant configuration values.

Add more logging, consistently, including relevant configuration values in log entry.

## How can you test this?

Tests + staging deploy

### DIFF
--- a/step_start_instance.go
+++ b/step_start_instance.go
@@ -33,9 +33,6 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 
 	instance, err := s.provider.Start(ctx, buildJob.StartAttributes())
 	if err != nil {
-		logger.WithField("err", err).Error("couldn't start instance")
-		context.CaptureError(ctx, err)
-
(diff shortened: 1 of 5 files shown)


step_start_instance.go
```diff
@@ -33,9 +33,6 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 
 	instance, err := s.provider.Start(ctx, buildJob.StartAttributes())
 	if err != nil {
-		logger.WithField("err", err).Error("couldn't start instance")
```
soulshake: What's the reasoning behind moving this log entry down? Won't this error get swallowed if we return at line 49 (`return multistep.ActionHalt`) ?

meatballhat: The error condition handled in that block is specifically for `workererrors.JobAbortError`, which is not an internal error that should necessarily be logged as such.  Moving this code also makes it more consistent with other places where we run `buildJob.Requeue`.

PR #419, opened 2017-11-02
Add INSPECT_INTERVAL to customize the duration between container inspections

soulshake:

Currently, travis-worker inspects containers every 500 milliseconds, and this is not configurable. This may be contributing to the docker stalls described in https://github.com/travis-pro/team-blue/issues/704.

This PR adds an environment variable `TRAVIS_WORKER_DOCKER_INSPECT_INTERVAL` where this value can be defined (in milliseconds).

### The easy way

`$ go test -v ./backend -test.run InspectInterval`

### The hard way

Set `"debug": true` in `/etc/docker/daemon.json` and run `sudo kill -s SIGHUP $(pidof dockerd)` to reload its configuration.

```
$ make build
$ export TRAVIS_WORKER_DOCKER_INSPECT_INTERVAL=1500
$ travis-worker
```

Enqueue some jobs via the file queue, and observe the duration between container inspections in the docker debug log.

Tell me everything <3

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -170,11 +170,11 @@ func newDockerProvider(cfg *config.ProviderConfig) (Provider, error) {
 
 	inspectInterval := defaultInspectInterval
 	if cfg.IsSet("INSPECT_INTERVAL") {
-		v, err := strconv.ParseInt(cfg.Get("INSPECT_INTERVAL"), 10, 64)
+		v, err := time.ParseDuration(cfg.Get("INSPECT_INTERVAL"))
 		if err != nil {
 			return nil, err
 		}
-		inspectInterval = time.Duration(v) * time.Millisecond
+		inspectInterval = v
 	}
 
 	binds := []string{}
@@ -250,7 +250,7 @@ func newDockerProvider(cfg *config.ProviderConfig) (Provider, error) {
 		imageSelector: imageSelector,
 
 		execCmd:         execCmd,
-		inspectInterval: time.Duration(inspectInterval),
+		inspectInterval: inspectInterval,
 		tmpFs:           tmpFs,
 
 		cpuSets: make([]bool, cpuSetSize),
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -51,7 +51,7 @@ var (
 		"CERT_PATH":           "directory where ca.pem, cert.pem, and key.pem are located (default \"\")",
 		"CMD":                 "command (CMD) to run when creating containers (default \"/sbin/init\")",
 		"EXEC_CMD":            fmt.Sprintf("command to run via exec/ssh (default %q)", defaultExecCmd),
-		"INSPECT_INTERVAL":    fmt.Sprintf("time to wait between container inspections in ms (default %q)", defaultInspectInterval),
+		"INSPECT_INTERVAL":    fmt.Sprintf("time to wait between container inspections as duration (default %q)", defaultInspectInterval),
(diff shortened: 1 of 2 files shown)


backend/docker.go
```diff
@@ -165,6 +168,15 @@ func newDockerProvider(cfg *config.ProviderConfig) (Provider, error) {
 		execCmd = strings.Split(cfg.Get("EXEC_CMD"), " ")
 	}
 
+	inspectInterval := defaultInspectInterval
+	if cfg.IsSet("INSPECT_INTERVAL") {
+		v, err := strconv.ParseInt(cfg.Get("INSPECT_INTERVAL"), 10, 64)
```
meatballhat: I would prefer using `time.ParseDuration` instead, which means we don't have to assume milliseconds.

backend/docker.go
```diff
@@ -237,8 +249,9 @@ func newDockerProvider(cfg *config.ProviderConfig) (Provider, error) {
 		runNative:     runNative,
 		imageSelector: imageSelector,
 
-		execCmd: execCmd,
-		tmpFs:   tmpFs,
+		execCmd:         execCmd,
+		inspectInterval: time.Duration(inspectInterval),
```
meatballhat: Did you have to explicitly cast this value to a duration?

backend/docker_test.go
```diff
@@ -339,6 +339,27 @@ func TestNewDockerProvider_WithCMD(t *testing.T) {
 	assert.Equal(t, []string{"/bin/bash", "/fancy-docker-init-thing"}, provider.runCmd)
 }
 
+func TestNewDockerProvider_WithInspectInterval(t *testing.T) {
+	provider, err := dockerTestSetup(t, config.ProviderConfigFromMap(map[string]string{
+		"INSPECT_INTERVAL": "1500",
```
meatballhat: With `time.ParseDuration`, this value would instead be `1500ms` (or `1.5s`).

backend/docker.go
```diff
@@ -43,13 +43,15 @@ const (
...
 	defaultDockerSSHDialTimeout                = 5 * time.Second
+	defaultInspectInterval                     = 500 * time.Millisecond
 	defaultExecCmd                             = "bash /home/travis/build.sh"
 	defaultTmpfsMap                            = map[string]string{"/run": "rw,nosuid,nodev,exec,noatime,size=65536k"}
 	dockerHelp                                 = map[string]string{
 		"ENDPOINT / HOST":     "[REQUIRED] tcp or unix address for connecting to Docker",
 		"CERT_PATH":           "directory where ca.pem, cert.pem, and key.pem are located (default \"\")",
 		"CMD":                 "command (CMD) to run when creating containers (default \"/sbin/init\")",
 		"EXEC_CMD":            fmt.Sprintf("command to run via exec/ssh (default %q)", defaultExecCmd),
+		"INSPECT_INTERVAL":    fmt.Sprintf("time to wait between container inspections in ms (default %q)", defaultInspectInterval),
```
meatballhat: "time to wait between container inspections as duration" ?

backend/docker_test.go
```diff
@@ -339,6 +339,27 @@ func TestNewDockerProvider_WithCMD(t *testing.T) {
...
+}
+
+func TestNewDockerProvider_WithInvalidInspectInterval(t *testing.T) {
+	provider, err := dockerTestSetup(t, config.ProviderConfigFromMap(map[string]string{
+		"INSPECT_INTERVAL": "mraaaaaaa",
+	}))
+	defer dockerTestTeardown()
+
+	assert.NotNil(t, err)
+	assert.Equal(t, "time: invalid duration mraaaaaaa", err.Error())
```
meatballhat: 😻

PR #420, opened 2017-11-03
Experimental Librato metrics library swap

meatballhat:

The metrics libraries we currently use are reporting metrics that sometimes (always?) do not match up with log output, such as the requeue counts reported on GCE workers in the hundreds/min while logs show tens/min.

Use the Librato client library recommended by Librato, plus some request aggregation code.

## How can you test this?

- [ ] canary production deployment

### DIFF
--- a/amqp_job.go
+++ b/amqp_job.go
@@ -75,7 +75,7 @@ func (j *amqpJob) Received(ctx gocontext.Context) error {
 	j.received = time.Now()
 
 	if j.payload.Job.QueuedAt != nil {
-		metrics.TimeSince("travis.worker.job.queue_time", *j.payload.Job.QueuedAt)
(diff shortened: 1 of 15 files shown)


meatballhat:
😬 
```
time="2017-11-04T01:03:21Z" level=info msg="starting librato metrics reporter" pid=1 self=cli
panic: interface conversion: interface {} is uint64, not int64

goroutine 11 [running]:
github.com/travis-ci/worker/metrics.ReportMemstatsMetrics()
        /go/src/github.com/travis-ci/worker/metrics/memstats.go:32 +0x9fc
created by github.com/travis-ci/worker.(*CLI).setupMetrics
        /go/src/github.com/travis-ci/worker/cli.go:331 +0x210

```

meatballhat:
Maybe another time!

PR #422, opened 2017-11-07
Disable docker cpuset allocation when CPUS is 0

meatballhat:
as advertised in the help text, but not currently true

The help text claims that setting `CPUS` to 0 will disable cpuset allocation, but this is not currently true.

Do like the help text claims.

## How can you test this?

- [x] staging deployment with `export TRAVIS_WORKER_DOCKER_CPUS=0`

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -90,7 +90,7 @@ type dockerProvider struct {
 	runBinds        []string
 	runMemory       uint64
 	runShm          uint64
-	runCPUs         int
+	runCPUs         uint
 	runNative       bool
 	execCmd         []string
 	inspectInterval time.Duration
@@ -245,7 +245,7 @@ func newDockerProvider(cfg *config.ProviderConfig) (Provider, error) {
 		runBinds:      binds,
 		runMemory:     memory,
 		runShm:        shm,
-		runCPUs:       int(cpus),
+		runCPUs:       uint(cpus),
 		runNative:     runNative,
 		imageSelector: imageSelector,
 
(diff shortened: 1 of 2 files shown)


backend/docker.go
```diff
@@ -405,18 +405,23 @@ func (p *dockerProvider) Start(ctx gocontext.Context, startAttributes *StartAttr
...
-			"err":            err,
-			"cpu_set_length": len(p.cpuSets),
-			"run_cpus":       p.runCPUs,
-		}).Error("couldn't checkout CPUSets")
-		return nil, err
-	}
+	useCPUSets := p.runCPUs != uint(0)
+	cpuSets := ""
+
+	if useCPUSets {
```
meatballhat: @soulshake This guard condition ensures that the CPU set checkout never happens when using CPU shares, which also means that the CPU set size isn't checked.  Does that help explain behavior?

meatballhat:
@soulshake There are tests that cover the defaulted CPU set size of 2, fwiw 👍

backend/docker.go
soulshake: :+1:

soulshake:
@meatballhat Sorry, I meant a test for when the CPU size is 0 :)

meatballhat:
@soulshake do you mean like this one? https://github.com/travis-ci/worker/blob/edb4401e7ac15c73fce575e46cf196a26f2f8731/backend/docker_test.go#L321-L330

soulshake:
@meatballhat Ah yes, that seems like it'll jolly well do :+1:

PR #423, opened 2017-11-08
Prepare for v3.4.0 release

meatballhat:

The size of [this diff](https://github.com/travis-ci/worker/compare/v3.3.1...HEAD).

Prepare for v3.4.0 release given changes and additions beyond a patch level bump and no breaking changes to public APIs.

## How can you test this?

- [x] production canary deployment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ## [Unreleased]
 
 ### Added
-- backend/docker: add repo, job ID, and dist as container labels
 
 ### Changed
-- http-job-queue: use polling and refresh claim intervals from job delivery
-  responses, if available
 
 ### Deprecated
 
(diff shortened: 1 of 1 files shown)


bnferguson:
Digging into the change log I see: `backend/docker: add repo, job ID, and dist as container labels`

Where can I find more information about what this is, and how to use it? If I'm reading this right, does that mean we can have precise, trusty, and eventually xenial build images all on the same worker even when we're not using job board/jupiterbrain? (aka selecting via docker labels). 

That'd be a huuuge game changer for us in Enterprise with the trusty roll out.

meatballhat:
@bnferguson Hello!  The container labels are additional container metadata, and sadly aren't related to image selection.  I would also love to see an image selection option that doesn't rely on job-board allowing dist-specific selection, so I'll work up a PR shortly ❤️

meatballhat:
@bnferguson here goes! https://github.com/travis-ci/worker/pull/424

bnferguson:
omg, you weren't kidding about shortly! 💖

PR #424, opened 2017-11-09
Add support for env-based image selection to Docker backend

meatballhat:

Image selection in the Docker backend is currently limited to tag-based and api-based strategies, which means that the only "local" strategy (tag-based) will only allow for selection based on `language`, which means that multiple distro images cannot be used in the same Docker API.

Add support for env-based image selection.

## How can you test this?

The env-based image selector itself is well-tested.  Testing this in a canary deployment would be nice, too!

- [x] add docs to README about how to configure env-based image selection
- [x] does it work?

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,42 @@ built-in help system:
 travis-worker --help
 ```
 
+### Environment-based image selection configuration
+
+Some backend providers supports image selection based on environment variables.
+The required format uses keys that are prefixed with the provider-specific
+prefix:
+
+- `TRAVIS_WORKER_{UPPERCASE_PROVIDER}_IMAGE_ALIASES`: comma-delimited strings used for
+  looking up full image name strings from other environment variables
+- `TRAVIS_WORKER_{UPPERCASE_PROVIDER}_IMAGE_ALIAS_{UPPERCASE_ALIAS}`: contains
+  an alias name which will be used to look up another environment variable, such
+as when routing multiple languages to the same image
+- `TRAVIS_WORKER_{UPPERCASE_PROVIDER}_IMAGE_{UPPERCASE_ALIAS}`: contains a full image name
+  string to be used by the backend provider
+
+The following example is for use with the Docker backend:
+
+``` bash
+# declare aliases for `dist: trusty`, `os: linux`, and `group: dev`
+export TRAVIS_WORKER_DOCKER_IMAGE_ALIASES=dist_trusty,os_linux,group_dev
+export TRAVIS_WORKER_DOCKER_IMAGE_ALIAS_GROUP_DEV=bionic
+export TRAVIS_WORKER_DOCKER_IMAGE_ALIAS_DIST_TRUSTY=trusty
+export TRAVIS_WORKER_DOCKER_IMAGE_ALIAS_OS_LINUX=trusty
+
+# resolves the above alias of `trusty`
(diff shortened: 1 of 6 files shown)


bnferguson:
THE GAME HAS CHANGED

!!!!!!!!!!!!!!!!!!!!!

This is amazing! Reviewing now!

backend/docker.go
```diff
@@ -304,6 +300,8 @@ func buildDockerImageSelector(selectorType string, client *docker.Client, cfg *c
 	switch selectorType {
 	case "tag":
 		return &dockerTagImageSelector{client: client}, nil
+	case "env":
+		return image.NewEnvSelector(cfg)
```
bnferguson: Wait, like, you weren't kidding about this already existing!

So how does this play out for which dockerTag to snag? Or is it by another means entirely? I'm guessing this will require some updates to the worker config and install script?

meatballhat: It's completely separate from docker tags, yep!  I would like to include some env-based selection docs in this PR, too.  Once upon a time, I thought we had some docs about it, but I don't see any in the README, so fixing!

README.md
```diff
@@ -48,6 +48,42 @@ built-in help system:
...
+# resolves the above alias of `bionic`
+export TRAVIS_WORKER_DOCKER_IMAGE_BIONIC=registry.business.com/fancy/ubuntu:bionic
+
+# resolves as the fallback default image
+export TRAVIS_WORKER_DOCKER_IMAGE_DEFAULT=travisci/ci-garnet:packer-1410230255-fafafaf
+
+# resolves for `language: ruby`
+export TRAVIS_WORKER_DOCKER_IMAGE_LANGUAGE_RUBY=registry.business.com/travisci/ci-ruby:whatever
+```
+
```
bnferguson: This is really useful! So to use this back end and these env vars, we just need to use `IMAGE_SELECTOR_TYPE='env'`? as part of the worker config? 

👍 From me!

meatballhat: @bnferguson The full env var to enable this would be either of these, but otherwise yes!:
- `TRAVIS_WORKER_DOCKER_IMAGE_SELECTOR_TYPE=env`
- `DOCKER_IMAGE_SELECTOR_TYPE=env`

meatballhat:
@bnferguson In short, this implementation is somewhere between busted and confusing to the point of being unusable &lt;/subjectivity&gt;  I would like to work on an alternate implementation now that (AFAIK) there are no production instances relying on env-based image selection.

README.md
```diff
@@ -60,6 +60,34 @@ built-in help system:
...
+# resolves for `language: ruby`
+export TRAVIS_WORKER_DOCKER_IMAGE_RUBY=registry.business.com/travisci/ci-ruby:whatever
+
+# resolves for `group: edge` + `language: python`
+export TRAVIS_WORKER_DOCKER_IMAGE_GROUP_EDGE_PYTHON=travisci/ci-garnet:packer-1530230255-fafafaf
+
+# used when no dist, language, or group matches
+export TRAVIS_WORKER_DOCKER_IMAGE_DEFAULT=travisci/ci-garnet:packer-1410230255-fafafaf
+```
+
```
bnferguson: Oh nice. This is super doable from replicated even once we manage things there. 👍 This feels super clear how we might do this, and even update installs to newer images. Also feels flexible for our more advanced customers that do some customizations for specific setups (and use groups). 

This is great!

PR #432, opened 2018-02-06
[WIP] Send worker events to Honeycomb

bogdanap:
[Honeycomb](https://honeycomb.io/) is a pretty cool tool (see [Igor's blogpost](https://builders.travis-ci.com/blog/2017-10-18-honeycomb/)) and we can always use more info about what our systems are doing.

Hook worker up to Honeycomb. I've limited this PR to just one random event, since it's just meant to be the initial set-up phase. 

This will need keychain/terraform changes too (for the configuration bits), but it shouldn't fail without them.

## How can you test this?
Since this is just the set-up and experimentation, I haven't added any unit-tests, although I think they are useful on the long run. (I will probably add them before this PR is merged if we decide to go the Honeycomb route)
- [ ] Deploy to staging and check that data flows into Honeycomb
- [ ] Deploy to production and check that data flows into Honeycomb

Honeycomb sends the events asynchronously and uses a separate channel to return the responses (see [docs](https://honeycomb.io/docs/connect/go/#handling-responses)). I left response handling out for now to keep things simpler, but I think we do eventually want to check whether we're having issues sending events, right?  

Any other comments are of course welcome!

### DIFF
--- a/step_start_instance.go
+++ b/step_start_instance.go
@@ -68,10 +68,11 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 	}).Info("started instance")
 
 	libhoney.SendNow(map[string]interface{}{
+		"provider":    s.provider,
+		"job_id":      buildJob.Payload().Job.ID,
 		"boot_time":   time.Since(startTime),
 		"instance_id": instance.ID(),
 		"image_name":  instance.ImageName(),
-		"provider":    s.provider,
 	})
 
 	state.Put("instance", instance)
--- a/step_start_instance.go
+++ b/step_start_instance.go
@@ -71,7 +71,7 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 		"boot_time":   time.Since(startTime),
 		"instance_id": instance.ID(),
 		"image_name":  instance.ImageName(),
-		"provider": s.provider
+		"provider":    s.provider,
 	})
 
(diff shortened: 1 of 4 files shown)


bnferguson:
Just jumping in here to confirm that this will only be turned on if configured? As in Enterprise we won't want to have it be required. Looking at the code this seems to be the case but wanted to double check.

bogdanap:
@bnferguson Yes, it should only send metrics when configured and the configuration should be optional.

step_start_instance.go
```diff
@@ -66,6 +67,14 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 		"image_name":  instance.ImageName(),
 	}).Info("started instance")
 
+	libhoney.SendNow(map[string]interface{}{
+		"provider":    s.provider,
+		"job_id":      buildJob.Payload().Job.ID,
+		"boot_time":   time.Since(startTime),
+		"instance_id": instance.ID(),
+		"image_name":  instance.ImageName(),
+	})
```
igorwwwwwwwwwwwwwwwwwwww: perhaps we should also add a field for the name of the event, e.g. `"event": "started instance"` or such? not super important for a first test, but if we want to add more events we'll want to differentiate.

bogdanap:
We've decided that using [honeytail](https://honeycomb.io/docs/connect/agent/) and the log output is a better approach.

PR #437, opened 2018-02-16
Add more details about our current release process

bogdanap:
Update the release process, per @meatballhat suggestion in https://github.com/travis-pro/manual/pull/671#discussion_r168768865

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ Once you've decided what the next version number should be, update the [changelo
 Once the changelog has been updated and merged to `master`, the merge commit needs to be signed and manually tagged with the version number. To do this, run:
 
 ```
-$ git tag -a vX.X.X -m "Worker version vX.X.X"
+$ git tag --sign -a vX.X.X -m "Worker version vX.X.X"
 $ git push origin vX.X.X
 ```
 
--- a/README.md
+++ b/README.md
@@ -195,14 +195,22 @@ To add a new dependency, do the following:
 
 ## Release process
 
-The parts of the release process that haven't yet been automated look like this:
+Since we want to easily keep track of worker changes, we often associate them with a version number.
+To find out the current version, check the [changelog](https://github.com/travis-ci/worker/blob/master/CHANGELOG.md) or run `travis-worker --version`.
+We typically use [semantic versioning](https://semver.org/) to determine how to increase this number.
 
-- [ ] review the diff since last release for silliness
-- [ ] decide what the version bump should be
-- [ ] update [`./CHANGELOG.md`](./CHANGELOG.md) (in a release prep branch)
-- [ ] tag accordingly after merge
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,14 +195,22 @@ To add a new dependency, do the following:
...
+Since we want to easily keep track of worker changes, we often associate them with a version number.
+To find out the current version, check the [changelog](https://github.com/travis-ci/worker/blob/master/CHANGELOG.md) or run `travis-worker --version`.
+We typically use [semantic versioning](https://semver.org/) to determine how to increase this number.
+
+Once you've decided what the next version number should be, update the [changelog](https://github.com/travis-ci/worker/blob/master/CHANGELOG.md) making sure you include all relevant changes that happened since the previous version was tagged. You can see these by running `git diff vX.X.X...HEAD`, where `v.X.X.X` is the name of the previous version.
+
+Once the changelog has been updated and merged to `master`, the merge commit needs to be signed and manually tagged with the version number. To do this, run:
+
+```
+$ git tag -a vX.X.X -m "Worker version vX.X.X"
```
meatballhat: Can this recommend using `--sign`, also? 💖

README.md
```diff
@@ -195,14 +195,22 @@ To add a new dependency, do the following:
...
+$ git push origin vX.X.X
```
meatballhat: This may be specific to my current git config, but I have to run `git push --tags origin master` in order for the tags to push, fwiw.

bogdanap: I thought using `--tags` just pushes all new tags, as opposed to pushing specific ones by their names.

README.md
bogdanap: Added! Thanks!

README.md
meatballhat: @bogdanap that's correct, IIUC.  Given a situation where I have a clone of `travis-ci/worker` that is up to date with `origin`, and I create a new tag, then I want to push that tag to `origin`, then the only tag that I have locally that isn't already in `origin` will be pushed via `git push --tags origin master`.  I'm sure there is a syntax to push only a single tag, but I have never used it and my reading of `git-push(1)` doesn't leave me feeling any more confident 😅

bogdanap: @meatballhat I'm a bit confused - I thought `git push origin <tag>` at line `208` accomplishes just that - pushing the single tag, so the separate `push --tags` shouldn't be necessary. 😅 
I assumed that since v3.5.0 is visible to everyone, that works ok, but then again, I'm still getting familiar with `git` so 🤷‍♀️ .
 If you think adding the `push --tags ` is the better way to go, I'm happy to change it.

meatballhat: @bogdanap I'm so sorry.  I looked right at this line and in my head it was `git push origin master` 🤦‍♂️  Yes, this absolutely is a good way to push a tag.  Sorry again!  💖

PR #438, opened 2018-02-19
Waiting before running createdb fixes flaky integration test

bogdanap:
http-job-test is still failing quite often at the stage where it's trying to create the new db, but can't connect to postgres (example: https://travis-ci.org/travis-ci/worker/builds/342690250)

I thought using pg_isready would be a better way to check for postgres' status, but it seems we're still trying to connect a bit too fast (result: https://travis-ci.org/travis-ci/worker/builds/342718822).
I've decided to keep using it for the check (instead of the previous `psql -l`) since I feel it communicates the intention better.
Adding an extra sleep (as ugly as it is) between the check and the createdb call seems to improve things (haven't had any failures yet, but we can also considering increasing the duration if they start occurring again)

## How can you test this?
Check the builds for intermittent failures.

### DIFF
--- a/script/http-job-test
+++ b/script/http-job-test
@@ -22,7 +22,7 @@ main() {
 
   docker exec \
     --user postgres \
-    job-board-postgres bash -c 'while ! pg_isready; do sleep 1; done'
+    job-board-postgres bash -c 'while ! pg_isready -d template1; do sleep 1; done'
 
   docker exec \
     --user postgres \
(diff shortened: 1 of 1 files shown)


script/http-job-test
```diff
@@ -22,7 +22,11 @@ main() {
 
   docker exec \
     --user postgres \
-    job-board-postgres bash -c 'while ! psql -l; do sleep 1; done'
+    job-board-postgres bash -c 'while ! pg_isready; do sleep 1; done'
```
igorwwwwwwwwwwwwwwwwwwww: 😻

meatballhat: TIL!

PR #444, opened 2018-03-08
Improve worker logs

bogdanap:
https://github.com/travis-ci/reliability/issues/6
Since we now have a [honeycomb dataset for worker](https://ui.honeycomb.io/travis-ci/datasets/worker), which is based on the logs, we should keep the information in the logs easily searchable and filterable. 

As described in the [issue](https://github.com/travis-ci/reliability/issues/6), this is meant to:
- report instance boot time in seconds (so it's easier to parse by honeycomb, which currently treats it as a string)
- adds the  job id and repository slug to some events that are logged using the processors context (like canceling and requeueing)
- tracks how long it takes for a processor to perform a graceful shutdown

## How can you test this?
- [x] deploy to staging and check the [worker-staging dataset](https://ui.honeycomb.io/travis-ci/datasets/worker-staging)

### DIFF
--- a/step_start_instance.go
+++ b/step_start_instance.go
@@ -61,10 +61,10 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 	}
 
 	logger.WithFields(logrus.Fields{
-		"boot_time":   time.Since(startTime).Seconds() * 1e3,
-		"instance_id": instance.ID(),
-		"image_name":  instance.ImageName(),
-		"version":     VersionString,
+		"boot_duration_ms": time.Since(startTime).Seconds() * 1e3,
+		"instance_id":      instance.ID(),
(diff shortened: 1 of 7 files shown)


step_start_instance.go
```diff
@@ -61,9 +61,10 @@ func (s *stepStartInstance) Run(state multistep.StateBag) multistep.StepAction {
 	}
 
 	logger.WithFields(logrus.Fields{
-		"boot_time":   time.Since(startTime),
+		"boot_time":   time.Since(startTime).Seconds() * 1e3,
```
igorwwwwwwwwwwwwwwwwwwww: perhaps rename this to `boot_duration_ms`?

igorwwwwwwwwwwwwwwwwwwww:
there's qa few more logrus lines that use `time.Since()`, those should be changed from interval to numeric also.

PR #445, opened 2018-03-08
Make str2map accept both comma- and space-delimited strings

soulshake:
Make str2map accept both comma- and space-delimited strings strings in `str2map` (used by e.g. `TMPFS_MAP` and `CONTAINER_LABELS`).

Worker help text claims to look for comma-delimited strings for `TMPFS_MAP` and `CONTAINER_LABELS`, but it's a lie.

Splitting only on spaces [is problematic](https://github.com/travis-ci/terraform-config/pull/373), because we use Docker's `--env-file` option which treats quotes as part of the value. 

Update the `str2map` function to split on both commas and spaces.

I have also added a test for `str2map`.

<strike>This could potentially be a breaking change for anyone specifying multiple values in `TMPFS_MAP`.

Should I make this change backwards-compatible by splitting on both spaces _and_ commas? Output a warning if `TMPFS_MAP` contains spaces?</strike> This change should be backwards-compatible.


backend/package.go
```diff
@@ -114,7 +114,7 @@ func asBool(s string) bool {
 func str2map(s string) map[string]string {
 	ret := map[string]string{}
 
-	for _, kv := range strings.Split(s, " ") {
+	for _, kv := range strings.Split(s, ",") {
```
meatballhat: Can this instead be a regexp split on `[ ,]`?  The `TMPFS_MAP` config option is a case that's already being used by some Travis Enterprise installations.  If we want to be really thorough, I think that each item in `kvParts` should also be URI-decoded :grin:

soulshake: @meatballhat :+1: see latest commits

backend/package.go
```diff
@@ -123,7 +128,7 @@ func str2map(s string) map[string]string {
 		if len(kvParts) == 1 {
 			ret[key] = ""
 		} else {
-			ret[key] = strings.TrimSpace(kvParts[1])
+			ret[key] = url.QueryEscape(strings.TrimSpace(kvParts[1]))
```
meatballhat: Ah, sorry!  I meant the inverse of this so that env vars may contain shell-unsafe chars or one of the split chars `[ ,:]`.

PR #446, opened 2018-03-26
Rename a struct private var to golang convention

meatballhat:


meatballhat:
@bogdanap :sparkling_heart:

PR #452, opened 2018-04-25
backend/gce: Pass context to all GCE API calls

sarahhodne:

Prevent stuck HTTP API calls from causing the worker to be stuck.

Pass in the existing context to cancel the API requests when the context gets cancelled.

Not so much feedback as suggestions for further improvements: You may want to set a shorter timeout just before the API call for the context (probably a couple of seconds at most, with a retry) instead of the much longer timeout that's already set.

Also, not sure if you will find this useful or not, but `instances.insert` takes a `requestId` UUID, if you send multiple requests with the same UUID, it will ignore the further requests and return the response for the original one, see [the API docs](https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert) for more info (search for `requestId`). So if there's a network request issue, and the request made it to the server, but you never get the response, you can do retries with the same request ID and prevent duplicate VMs. Writing the code for that was a bit more involved, though, so I decided not to unless this is an issue you see often enough to want to resolve.

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -543,7 +543,7 @@ func (p *gceProvider) Setup(ctx gocontext.Context) error {
 	var err error
 
 	p.apiRateLimit(ctx)
-	p.ic.Zone, err = p.client.Zones.Get(p.projectID, p.cfg.Get("ZONE")).Do()
+	p.ic.Zone, err = p.client.Zones.Get(p.projectID, p.cfg.Get("ZONE")).Context(ctx).Do()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Looks like `DOCKER_HUB_TRIGGER_URL` is a secret env var, so it is not set for third party PRs. Let me see if we can change that to an `exit 0` instead.

igorwwwwwwwwwwwwwwwwwwww:
Took a shot at fixing that here: https://github.com/travis-ci/worker/pull/453.

PR #454, opened 2018-05-15
rabbitmq state update channel pooling

igorwwwwwwwwwwwwwwwwwwww:

We are currently creating several channels _per processor_. This means that as we scale worker counts, the number of channels will go up rapidly (currently at ~16k channels on org).

We have observed that channel count has a direct effect on memory usage. Since memory is limited, channel count could become a scalability bottleneck.

By placing common tasks like status updates (and possibly log delivery) into a worker pool, we can scale it independently. The hope is that we don't need one channel per processor to get the job done (pun kinda intended).

## How can you test this?

I have been testing it locally via `script/drain-logs` and `script/publish-example-payload`, but we'll want to test this extensively on both staging and production if we decide to roll with it.

This is still early stages, so: Does this seem like a sensible approach?

EDIT: ready for more in-depth review now!

refs travis-ci/reliability#91


igorwwwwwwwwwwwwwwwwwwww:
I've tested forcefully shutting down rabbit, and that seems to be working. 👍

igorwwwwwwwwwwwwwwwwwwww:
testing this on staging now!

amqp_job.go
```diff
@@ -163,29 +164,86 @@ func (j *amqpJob) createStateUpdateBody(ctx gocontext.Context, state string) map
...
+	p := payload.(*amqpStateUpdatePayload)
+	ctx, cancel := gocontext.WithCancel(p.ctx)
+
+	w.ctx = ctx
+	w.cancel = cancel
+
+	return w.sendStateUpdate(p)
+}
+
+func (w *amqpStateUpdateWorker) BlockUntilReady() {
```
bogdanap: I had to stop and think for a bit why this doesn't do anything. I assume it's because the sendStateUpdate operation is already blocking, so we don't need it, but it might be good to add a comment here (if that's correct).

igorwwwwwwwwwwwwwwwwwwww: that is correct, afaik `BlockUntilReady` exists for if you want to perform some background or warm-up work before accepting the next job, and ensures that the worker does not receive any jobs until then. i'll add a comment to clarify! :)

igorwwwwwwwwwwwwwwwwwwww:
looks solid on staging, i'll go ahead and merge this, and plan on canarying it in production next week!

PR #459, opened 2018-06-11
Add the concept of a separate LogsQueue 

bogdanap:
Reference issue: https://github.com/travis-ci/reliability/issues/90
This is the first step in separating the logs behaviour in general and allowing worker to use a separate RabbitMQ cluster for build log processing. 
For now, the worker just connects to a new AMQP server if given the option. 

I intentionally chose to create a new interface instead of reusing the LogWriter one in order to keep the current behaviour and transition the responsibility gradually. 
For now, this only applies to AMQP queues (and not the HTTP or file types), but in the future it should be easy to switch between queue types separately for jobs and logs.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -211,8 +211,8 @@ func (i *CLI) Run() {
 	go i.signalHandler()
 
 	i.logger.WithFields(logrus.Fields{
-		"pool_size": i.Config.PoolSize,
-		"queue":     i.JobQueue,
+		"pool_size":  i.Config.PoolSize,
+		"queue":      i.JobQueue,
 		"logs_queue": i.LogsQueue,
 	}).Debug("running pool")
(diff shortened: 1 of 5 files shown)


cli.go
```diff
@@ -206,6 +213,7 @@ func (i *CLI) Run() {
 	i.logger.WithFields(logrus.Fields{
 		"pool_size": i.Config.PoolSize,
 		"queue":     i.JobQueue,
+		"log_queue": i.LogsQueue,
```
igorwwwwwwwwwwwwwwwwwwww: Maybe keep the terminology consistent here in the key? i.e. `logs_queue`

bogdanap:
This was tested both locally and on staging. ✅

PR #460, opened 2018-06-13
Make heartbeat configurable for AMQP connections

bogdanap:
https://github.com/travis-ci/reliability/issues/114

Set hearbeats to 0 in order to disable them (just for the insecure connections, for testing purposes - if this works, will add the same for the amqp tls connection as well).

## How can you test this?
Check for normal behaviour on staging.
Deploy to a few instances and see if errors like `[ERROR] closing AMQP connection <0.22954.214> (35.188.132.197:38448 -> 10.50.94.192:5671): missed heartbeats from client, timeout: 10s
` stops appearing in the cloud amqp logs

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -627,7 +627,11 @@ func (i *CLI) buildAMQPJobQueueAndCanceller() (*AMQPJobQueue, *AMQPCanceller, er
 			&tls.Config{InsecureSkipVerify: true},
 		)
 	} else {
-		amqpConn, err = amqp.Dial(i.Config.AmqpURI)
+		amqpConfig := amqp.Config{
+			Heartbeat: 0 * time.Second,
+			Locale:    "en_US",
+		}
+		amqpConn, err = amqp.DialConfig(i.Config.AmqpURI, amqpConfig)
 	}
 	if err != nil {
 		i.logger.WithField("err", err).Error("couldn't connect to AMQP")
--- a/cli.go
+++ b/cli.go
@@ -620,18 +620,26 @@ func (i *CLI) buildAMQPJobQueueAndCanceller() (*AMQPJobQueue, *AMQPCanceller, er
 			}
 			cfg.RootCAs.AppendCertsFromPEM(cert)
 		}
-		amqpConn, err = amqp.DialTLS(i.Config.AmqpURI, cfg)
+		amqpConn, err = amqp.DialConfig(i.Config.AmqpURI,
+			amqp.Config{
+				Heartbeat:       i.Config.AmqpHeartbeat,
+				Locale:          "en_US",
+				TLSClientConfig: cfg,
+			})
 	} else if i.Config.AmqpInsecure {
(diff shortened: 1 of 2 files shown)


bogdanap:
> Would it make sense to make this configurable?
Considering the fact that we haven't really used this option before and that worker config is already pretty complex, I'd rather just rely on keeping either on/off by default (or until the next time we need to change it).

igorwwwwwwwwwwwwwwwwwwww:
The reason I asked is to make it easier to revert if it causes any issues. Rolling back worker also works for me though! 👍

bogdanap:
This ran without errors on GCE staging. It's hard, however, to tell if it helps with the heartbeat errors since we don't get them in staging, so I'll move on with canary-ing it in production

igorwwwwwwwwwwwwwwwwwwww:
It might be a fallback behaviour in the client library, I found this ticket: https://github.com/streadway/amqp/pull/329. We can try and get some movement on that PR, but it probably makes sense to try increasing the timeout in the mean time.

config/config.go
```diff
@@ -50,6 +50,10 @@ var (
 			Value: defaultQueueType,
 			Usage: `The name of the queue type to use ("amqp", "http", or "file")`,
 		}),
+		NewConfigDef("AmqpHeartbeat", &cli.DurationFlag{
+			Value: 10 * time.Second,
+			Usage: "The heartbeat timeout value defines after what time the peer TCP connection should be considered unreachable",
```
igorwwwwwwwwwwwwwwwwwwww: This is a bit confusing, since the heartbeat is an AMQP-level concept and is only indirectly related to TCP. AFAIK triggering periodic traffic on an idle connection imitates TCP keepalive and allows clients to detect a broken connection.

cli.go
```diff
@@ -620,14 +620,26 @@ func (i *CLI) buildAMQPJobQueueAndCanceller() (*AMQPJobQueue, *AMQPCanceller, er
 			}
 			cfg.RootCAs.AppendCertsFromPEM(cert)
 		}
-		amqpConn, err = amqp.DialTLS(i.Config.AmqpURI, cfg)
+		amqpConn, err = amqp.DialConfig(i.Config.AmqpURI,
+			amqp.Config{
+				Heartbeat:       i.Config.AmqpHeartbeat,
+				Locale:          "en_US",
```
igorwwwwwwwwwwwwwwwwwwww: Is the addition of `Locale` something that we needed to add? Mostly curious.

config/config.go
bogdanap: This is the definition from the docs: https://www.rabbitmq.com/heartbeats.html#heartbeats-timeout. 
Do you have any suggestions on how to make it less confusing? 😅

igorwwwwwwwwwwwwwwwwwwww: Fair enough! :)

bogdanap:
Re-tested on staging! ✅

PR #462, opened 2018-06-21
Add support for setting RabbitMQ consumer priority

meatballhat:
defaulting to 0, which is the same as the RabbitMQ default.

My intent here is to ensure that given a travis-worker process with priority `> 0`, that it will be fully utilized prior to other travis-worker processes with `priority == 0`, e.g.:

- `travis-worker` `0` on lower-cost reserved capacity: `priority = 1`
- `travis-worker` `1` on higher-cost dynamic capacity: `priority = 0` (unset)

- [x] works correctly in production

### DIFF
--- a/amqp_job_queue.go
+++ b/amqp_job_queue.go
@@ -18,8 +18,9 @@ import (
 
 // AMQPJobQueue is a JobQueue that uses AMQP
 type AMQPJobQueue struct {
-	conn  *amqp.Connection
-	queue string
+	conn     *amqp.Connection
+	queue    string
+	priority int
 
 	stateUpdatePool *tunny.Pool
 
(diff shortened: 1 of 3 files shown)


meatballhat:
I will be testing this out on a production canary server shortly and will report back :sparkles:

meatballhat:
:scream_cat: 

```
time="2018-06-21T14:22:50Z" level=error msg="couldn't create processor" err="table field \"x-priority\" value %!t(int=1) not supported" pid=1 self=processor_pool
```

meatballhat:
I guess I'll go with `int16` (???)  https://github.com/streadway/amqp/blob/fc7fda2371f5327ad39211e09482845b8734cc72/types.go#L228

meatballhat:
Looking good now!

rabbitmq-consumer-args-x-priority

PR #464, opened 2018-06-21
Enable the use of the RabbitMQ sharding plugin for travis-logs

bogdanap:
https://github.com/travis-ci/reliability/issues/125

I tried to follow the instructions from the [README of the plugin](https://github.com/rabbitmq/rabbitmq-sharding).
I've declared a new exchange (`reporting.jobs.logs_sharded`, to match what the [changes in the logs app](https://github.com/travis-ci/travis-logs/pull/181) expected and made sure the `LogWriter` routes to the correct exchange based on whether the new config option `TRAVIS_WORKER_RABBITMQ_SHARDING` is set or not.
In terms of manual changes, I had to enable the RabbitMQ sharding plugin in the CloudAMQP Plugins tab and create a new policy (`logs-sharding`) for the sharded exchage, as specified in the README that's linked above.

(Naming could probably be improved, since I was mostly interested in getting this in a state where I could test it before leaving for the weekend 😅)

## How can you test this?
This was successfully tested on org-staging. The CloudAMQP addon has been configured as described above (with the plugin enabled and exchange policy created) and all org workers are currently running in sharded mode (the changes were made locally on the instances and not through a `terraform apply`)
I've also tested with just half of the workers running in sharded mode and without the logs drain_sharded to see the messages piling up in the sharded queues. 
Everything looked good! ✅

### DIFF
--- a/amqp_job_queue.go
+++ b/amqp_job_queue.go
@@ -30,7 +30,7 @@ type AMQPJobQueue struct {
 // connects to the AMQP queue with the given name. The queue will be declared
 // in AMQP when this function is called, so an error could be raised if the
 // queue already exists, but with different attributes than we expect.
-func NewAMQPJobQueue(conn *amqp.Connection, queue string, stateUpdatePoolSize int) (*AMQPJobQueue, error) {
+func NewAMQPJobQueue(conn *amqp.Connection, queue string, stateUpdatePoolSize int, sharded bool) (*AMQPJobQueue, error) {
 	channel, err := conn.Channel()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


amqp_job_queue.go
```diff
@@ -42,24 +43,32 @@ func NewAMQPJobQueue(conn *amqp.Connection, queue string, stateUpdatePoolSize in
...
 	if err != nil {
 		return nil, err
 	}
 
-	err = channel.QueueBind("reporting.jobs.logs", "reporting.jobs.logs", "reporting", false, nil)
-	if err != nil {
-		return nil, err
+	if sharded {
+		// This exchange should be declared as sharded using a policy that matches its name.
+		err = channel.ExchangeDeclare("reporting.jobs.logs_sharded", "direct", true, false, false, false, nil)
```
igorwwwwwwwwwwwwwwwwwwww: I believe we may need to use a `x-modulus-hash` exchange instead of a direct exchange, [as suggested by rabbitmq-sharding](https://github.com/rabbitmq/rabbitmq-sharding#messages-distribution-between-shards-partitioning).

The reason is kind of subtle and it was frankly hard to find in the rabbit docs. In [the routing tutorial](https://www.rabbitmq.com/tutorials/tutorial-four-python.html) under "Multiple bindings", they say:

> It is perfectly legal to bind multiple queues with the same binding key. In our example we could add a binding between X and Q1 with binding key `black`. In that case, the direct exchange will behave like fanout and will broadcast the message to all the matching queues.

So I believe that if we use a direct exchange with the sharding plugin, the same message will get copied to each of the shards, instead of distributed between them.

That's what I saw in a test I ran, but maybe you can see if you can verify that behaviour. 👍 

Side-note: If we _do_ go with the `x-modulus-hash` exchange, we'll need to define a routing key when publishing, which will used for hash partitioning.

bogdanap: Yep, this is indeed the case. Thank you for the suggestion!

PR #466, opened 2018-06-29
Allow custom VM config in .travis.yml for sudo-enabled jobs

soulshake:

This PR is intended to allow usage of GPUs on the GCE (sudo-enabled) provider.

The proposed change introduces a top-level `vm_config` key in .travis.yml, e.g. 
```
   "vm_config":{
      "gpu_count":1,
      "gpu_type":"nvidia-tesla-v100",
      "zone":"us-central1-a"
   },
```

Allowing the zone to be configured this way is needed* because some GPU types only exist in `us-central1-a`, others only exist in `us-central1-c`, etc. ([more info](https://cloud.google.com/compute/docs/gpus/#gpus-list))

 \*Alternatively, we could simply accept a `gpu_type` and then query to select a zone automatically. However, I'd be in favor of at least considering allowing this to be set explicitly for other (non-GPU-related) use cases.

## How can you test this?

You can use `example-payload-premium.json`.

Note: I needed to tweak `cli.go` to avoid getting timeouts on instance creation:
```
StartupTimeout:          time.Duration(20 * time.Second),
```

### "Machine type specified 'xxx' is in a different scope than the instance"

This works as expected when I use `us-central1-a`, but for some reason, trying to create job instances in `us-central1-c` fails with the following error:
```
err="googleapi: Error 400: Invalid value for field 'resource.machineType': 'https://www.googleapis.com/compute/v1/projects/travis-staging-1/zones/us-central1-c/machineTypes/n1-standard-4'. Machine type specified 'n1-standard-4' is in a different scope than the instance., invalid"
```
(A similar error happens with `n1-standard-2` as well.)

I was getting similar errors previously and came to the conclusion that it was being caused by things that must be in the same zone being in different zones (e.g. `AcceleratorType` and `MachineType` having URLs referring to different zones). But now everything should be in the same zone, and I'm still getting the error whenever I try to use `us-central1-c`. Could this be related to the `Disks.InitializeParams.SourceImage (`https://www.googleapis.com/compute/v1/projects/eco-emissary-99515/global/images/travis-ci-garnet-trusty-1510233122-5f9e53c`) which has a `sourceDisk` in `us-central1-a` (`projects/eco-emissary-99515/zones/us-central1-a/disks/packer-5a045423-5027-adb9-75be-557be077515d`)?

### Allowing provider-level defaults for `GPU_TYPE` and `GPU_COUNT`

I'm not sure this is really necessary! Should I take it out?


backend/gce.go
```diff
@@ -871,12 +908,60 @@ func buildGCEImageSelector(selectorType string, cfg *config.ProviderConfig) (ima
...
 	var machineType *compute.MachineType
 	switch startAttributes.VMType {
 	case "premium":
-		machineType = p.ic.PremiumMachineType
+		pic, err := p.client.MachineTypes.Get(p.projectID, zone.Name, p.cfg.Get("PREMIUM_MACHINE_TYPE")).Context(ctx).Do()
+		if err != nil {
+			logger.Warn(err)
+		}
+		machineType = pic
+	default:
```
meatballhat: With only two cases, I don't think this `switch` block is any better than an `if`, but given the API calls happening in each block I don't think it could be replaced with something like:

``` go
if someCondition {
    doSomething()
    return
}

doOtherThing()
```

or

``` go
value := someDefault

if someCondition {
    value = otherValue
}

File: backend/gce.go [hunk 2]
```diff
@@ -913,7 +998,13 @@ func (p *gceProvider) buildInstance(ctx gocontext.Context, startAttributes *Star
 	if p.deterministicHostname {
 		hostname = hostnameFromContext(ctx)
 	} else {
-		hostname = fmt.Sprintf("travis-job-%s", uuid.NewRandom())
+		hostname = fmt.Sprintf("aj-travis-job-%s", uuid.NewRandom())
```
meatballhat: I assume `aj-travis-job` is a dev leftover :sweat_smile:

backend/gce.go
```diff
@@ -913,7 +998,13 @@ func (p *gceProvider) buildInstance(ctx gocontext.Context, startAttributes *Star
...
+	}
+
+	acceleratorConfigs := []*compute.AcceleratorConfig{}
+	if acceleratorConfig.AcceleratorCount > 0 {
+		logger.Debug("GPU requested, setting acceleratorConfig")
+		acceleratorConfigs = []*compute.AcceleratorConfig{acceleratorConfig}
```
meatballhat: Append instead of reallocating?  &lt;/microoptimization&gt;

backend/gce.go
```diff
@@ -376,6 +383,22 @@ func newGCEProvider(cfg *config.ProviderConfig) (Provider, error) {
...
+	defaultAcceleratorConfig := &compute.AcceleratorConfig{}
+
+	defaultAcceleratorConfig.AcceleratorType = defaultGCEGpuType
+	if cfg.IsSet("GPU_TYPE") {
+		defaultAcceleratorConfig.AcceleratorType = cfg.Get("GPU_TYPE")
+	}
+
+	defaultAcceleratorConfig.AcceleratorCount = defaultGCEGpuCount
+	if cfg.IsSet("GPU_COUNT") {
+		dgc, err := strconv.ParseInt(cfg.Get("GPU_COUNT"), 0, 64)
```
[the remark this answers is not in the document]
soulshake: Might as well go with `int64` to match [compute.AcceleratorConfig](https://godoc.org/google.golang.org/api/compute/v1#AcceleratorConfig).

README.md
```diff
@@ -93,6 +93,10 @@ export TRAVIS_WORKER_BUILD_API_URI='https://x:API_KEY@build-staging.travis-ci.or
 `TRAVIS_WORKER_BUILD_API_URI` can be found in the env of the job board app, e.g.:
 `heroku config:get JOB_BOARD_BUILD_API_ORG_URL -a job-board-staging`.
 
+#### Images
+
+TODO
```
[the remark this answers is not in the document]
soulshake: To document these a little bit (with pointer to `--help`):

* `IMAGE_ALIASES`
* `IMAGE_DEFAULT `
* `IMAGE_PROJECT_ID`
* `IMAGE_SELECTOR_TYPE`
* `IMAGE_SELECTOR_URL`
* `IMAGE_[ALIAS_]{ALIAS}`

backend/gce.go
```diff
@@ -376,6 +383,22 @@ func newGCEProvider(cfg *config.ProviderConfig) (Provider, error) {
 		defaultImage = cfg.Get("IMAGE_DEFAULT")
 	}
 
+	defaultAcceleratorConfig := &compute.AcceleratorConfig{}
+
+	defaultAcceleratorConfig.AcceleratorType = defaultGCEGpuType
+	if cfg.IsSet("GPU_TYPE") {
```
soulshake: TODO: Add tests for this if we decide to keep it.

README.md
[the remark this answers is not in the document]
meatballhat: Ah, so at least some of these are specific to the "env" selector type, which used to be how the gce and jupiterbrain backends selected images in production.  Since the introduction of the "api" selector type, the env-based one has mostly fallen out of use on hosted Travis.  I found the placement in the `README.md` and section title confusing because I thought you were intending to address something at the top level for local development images, where I tend to think of image selection as being specific to each backend.  Sorry; rambling.   I guess I find the `Development: Running Travis Worker locally` to be confusing in that it is structured to describe steps for various backend providers, yet only the docker backend is filled in, and there's a docker-specific bit _outside_ of the docker-specific section ("Pull Docker images")...

meatballhat:
It does stuff!  https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/723813

PR #478, opened 2018-07-20
Next steps towards using a separate AMQP cluster for logs processing

bogdanap:
https://github.com/travis-ci/reliability/issues/90
Having a separate logs cluster is part of our strategy for RabbitMQ scalability.

Add the minimum necessary changes in order to get something functional and be able to test.
This makes things a bit confusing (with both the JobQueue and LogsQueue duplicating behaviour), but I expect some cleanup steps to follow.

## How can you test this?
Will try to test on staging. I'll consider things to be successful if there are messages being sent to the logs queue on the separate cluster.

ALL THE FEEDBACK!


bogdanap:
This is working on staging 🎉

cli.go
```diff
@@ -213,18 +213,18 @@ func (i *CLI) Run() {
 	i.logger.WithFields(logrus.Fields{
 		"pool_size":  i.Config.PoolSize,
 		"queue":      i.JobQueue,
-		"logs_queue": i.LogsQueue,
+		"logs_queue": i.LogWriterFactory,
```
igorwwwwwwwwwwwwwwwwwwww: Maybe change the key here as well?

bogdanap: ah, good catch, thanks! 😄

PR #481, opened 2018-08-07
Only terminate on host maintenance when GPUs are involved

soulshake:

Sudo-enabled job VMs are shutting down unexpectedly due to host maintenance events. (https://github.com/travis-ci/travis-ci/issues/4924#issuecomment-408327917)

According to https://github.com/travis-ci/worker/commit/811106032d78b3bff03434ed431e3ac90542781d, migrating on host maintenance is problematic when GPUs are involved.
This PR sets host maintenance action based on the presence of at least one GPU.

What's the actual problem caused by migrating when GPUs are present? Is there a way we can make migration work?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Added
 - backend/gce: build.sh trace persisting
+- backend/gce: make migration behavior on host maintenance events conditional on GPU usage
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


cotsog:
Is there an ETA when this will be merged and deployed? Thanks!

joshk:
@soulshake can we get this deployed this week?

aidanhs:
Hi @soulshake! We (Rust infra team) were just wondering where this is up to?

soulshake:
@aidanhs I will try to get this released today.

soulshake:
Ran into a hiccup when testing this in staging, so rollout is a bit delayed. I will post an update later today.

soulshake:
@aidanhs Rollout is about halfway complete and should be finished in the next few hours.

aidanhs:
Looks like we haven't seen any failures since the rollout - thanks @soulshake!

PR #482, opened 2018-08-20
Ensure correct build machine SSH permissions

whereswaldon:

In our Travis Enterprise deployment, certain machine images provision with incorrect permissions on `/home/travis/.ssh` and `/home/travis/.ssh/authorized_keys`. These files/directories are owned by user `root` and group `root`. When the worker tries to SSH in, the SSH agent is unable to read the authorized keys. This makes it reject the worker's login attempt and stall the build. The build job then requeues every five minutes indefinitely, but encounters this problem each time.

Rather than chase flaws in specific machine images, this patch simply ensures that the permissions are correct on any build machine.

We added a step to the script that the worker uses to start the build machine. This step `chown`s the SSH directory for user `travis` to be owned by user `travis` and group `travis`. This resolves the problem in all known instances. It also has the advantage of being idempotent.

## How can you test this?

We have been unable to get public Travis to replicate the problem, but our current hypothesis is that it's a specific build VM image that has the problem, and we don't know the build constraints that will make public Travis try to use that particular build image. On our infrastructure, the following `.travis.yml` always causes the problem:

```yaml
# uses a supported language
language: nix
# this ensures that it isn't choking our production queues with requeues
group: test
# always succeed
script: true
```

That script pulls the `travis-ci-connie-trusty-1510244212-511bc25` image, which exhibits the problem.

Could this change (or an equivalent) be accepted into the official release? It has virtually no runtime penalty and guards against an arcane category of failure when provisioning the build machines.


backend/gce.go
```diff
@@ -109,6 +109,7 @@ var (
 cat > ~travis/.ssh/authorized_keys <<EOF
 {{ .SSHPubKey }}
 EOF
+sudo chown -R travis:travis ~travis/.ssh/
```
meatballhat: Given that this script is _supposedly_ executed by `root`, I think I would prefer a version that omits `sudo` so that we aren't baking in a requirement for `sudo` to be present.

whereswaldon: Thanks for looking at this! We will test to make sure that it consistently succeeds without `sudo` and update the PR accordingly.

whereswaldon:
@meatballhat Thanks for your comment earlier! I've verified that `sudo` was unnecessary. Could I get a re-review?

PR #483, opened 2018-08-21
Add config options for jupiter-brain VM config

mjm:

We want to be able to experiment with alternate VM configurations for the Mac infra by changing the CPU count and RAM amount.

Added two new config options for the Jupiter Brain provider:

- `IMAGE_CPUS`: Number of CPUs to give the VM
- `IMAGE_RAM`: MB of RAM to give the VM

If the options are present, they get sent along to Jupiter Brain which will reconfigure the VM appropriately. If they are absent, the request to JB is just as it used to be.

## How can you test this?

Deployed this worker version to staging in pod-1 and ran some jobs. Looked at the VMs in vSphere and saw the correct configuration for the VMs. Builds also ran substantially faster than those in pod-2 which was unchanged.

Are there places where I should add tests? Does this overall approach seem okay?

### DIFF
--- a/backend/jupiterbrain.go
+++ b/backend/jupiterbrain.go
@@ -76,10 +76,10 @@ type jupiterBrainProvider struct {
 	bootPollDialTimeout  time.Duration
 	bootPollWaitForError time.Duration
 
-	imageSelectorType string
-	imageSelector     image.Selector
-	imageCPUs         int
-	imageRAM          int
+	imageSelectorType   string
+	imageSelector       image.Selector
+	defaultInstanceCPUs int
+	defaultInstanceRAM  int
 
 	apiClient *jupiterBrainAPIClient
 }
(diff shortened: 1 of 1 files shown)


backend/jupiterbrain.go
```diff
@@ -77,6 +78,8 @@ type jupiterBrainProvider struct {
 
 	imageSelectorType string
 	imageSelector     image.Selector
+	imageCPUs         int
+	imageRAM          int
```
igorwwwwwwwwwwwwwwwwwwww: Are CPU and RAM allocations a property of the image? The jupiter-brain patch read to me like they were instance-level settings.

FYI I know nothing about this. :)

mjm: I think the terms "instance" and "image" are roughly interchangeable here, at least the way we use them. The "images" we clone from are just shutdown virtual machines.

Maybe this is needlessly confusing. Would it make more sense if this was something like `defaultInstanceCPUs` and similar for RAM?

igorwwwwwwwwwwwwwwwwwwww: That makes sense, thanks for explaining! 👍

PR #486, opened 2018-08-23
only run trace code when trace flag is set on job

igorwwwwwwwwwwwwwwwwwwww:
This ensures a much stricter opt-in and should allow us to turn off tracing at the scheduler-level completely. That will enable us to perform a safer rollout of this feature.

https://github.com/travis-ci/reliability/issues/152


igorwwwwwwwwwwwwwwwwwwww:
Tested successfully on staging!

PR #492, opened 2018-08-29
End job when logs reach max size and cleanup instances after context cancelation

bogdanap:

https://github.com/travis-pro/team-blue/issues/501 and more specifically (https://github.com/travis-pro/team-blue/issues/501#issuecomment-416642948). 

In addition to that, I've noticed that when the context finishes due to a DeadlineExceeded error, the instance wasn't stopped - it always failed with a `context deadline exceeded error` (Example: `time="2018-08-29T17:21:30Z" level=warning msg="couldn't stop instance" err="context deadline exceeded" instance_id=travis-job-753e691c-02b1-4be8-9237-78940ed65cd0 job_id=422134858 job_path=firepress-org/ghostfire/jobs/422134858 pid=1 processor=f0b5ce82-85c1-40c0-9356-7dffd3c30551@1.production-2-worker-org-b-1-gce self=step_start_instance`)

In the first case, the job wasn't stopping after exceeding the log limit because the errors from the buffers Write function weren't handled as we'd expected (the SSH library was only checking those after the execution finished, which never happened, and we were simply ignoring the errors from the io.TeeReader in the native-docker case . This meant that we needed to find a different way to stop the job and I opted for a forced context cancellation, which seemed to do the trick.

In the second case, because the job context was created before we ran any of the job processing steps, it meant that some of the cleanup steps weren't available anymore if that context timed out or was otherwise prematurely canceled. Opting to reduce the scope and create a new derived context just for the RunScript step, solved the cleanup problem.

## How can you test this?
I tested both issues locally and it seemed to work.
Next step would be to deploy and test directly on staging.

Is this an ok approach? 
Could it potentially break other things?


amqp_log_writer.go
```diff
@@ -94,7 +95,9 @@ func (w *amqpLogWriter) Write(p []byte) (int, error) {
 		if err != nil {
 			logger.WithField("err", err).Error("couldn't write 'log length exceeded' error message to log")
 		}
-		return 0, ErrWrotePastMaxLogLength
+		gocontext.Canceled = ErrWrotePastMaxLogLength
```
igorwwwwwwwwwwwwwwwwwwww: This is probably going to break other things, since it replaces `Canceled` globally -- I believe _every_ canceled context will now be `ErrWrotePastMaxLogLength `. Not sure if there is any other way of setting a custom cancel error though.

bogdanap: That's a good point, I'm hoping https://github.com/travis-ci/worker/pull/492/commits/dee958c71883fcef51e9cfa00cea076e92484a85 should fix it.

amqp_log_writer.go
```diff
@@ -90,11 +92,10 @@ func (w *amqpLogWriter) Write(p []byte) (int, error) {
...
 	w.bytesWritten += len(p)
 	if w.bytesWritten > w.maxLength {
-		_, err := w.WriteAndClose([]byte(fmt.Sprintf("\n\nThe log length has exceeded the limit of %d MB (this usually means that the test suite is raising the same exception over and over).\n\nThe job has been terminated\n", w.maxLength/1000/1000)))
-		if err != nil {
-			logger.WithField("err", err).Error("couldn't write 'log length exceeded' error message to log")
-		}
-		return 0, ErrWrotePastMaxLogLength
+		logger.Info("wrote past maximum log length - cancelling context")
+		w.maxLengthReached = true
+		w.cancel()
```
igorwwwwwwwwwwwwwwwwwwww: Could there be cases where `w.cancel` is not set?

bogdanap: I think that if we care about termination for log length reasons, then it should be set. 
And since we can't properly process errors in this case, then crashing because it couldn't find the cancel func is the right thing to do - it will tell us that we've got something wrong. :)
Hopefully we'll catch these sorts of things in staging, because, despite what i said above, crashing in production is bad. 😅

igorwwwwwwwwwwwwwwwwwwww: While this would probably not happen in practice, we do write to the log before setting cancel, e.g. when worker writes the booting messages. If for some reason we write too much data there, we could run into a crash here.

Not likely in practice but imo it would be good to check for nil just in case.

bogdanap:
I've deployed this to GCE staging and intend to leave it running over the weekend to see how it behaves.

bogdanap:
This seems to be working without any issues, but I wanted to hold off deployment until I'll be back for good (after Monitorama and my long weekend next week) so I can keep an eye on production in case things go wrong.

PR #505, opened 2018-10-01
Implement downloading traces for Jupiter Brain

mjm:

We want to be able to use build tracing on the Mac infra, just like the other infras.

I copied the code from GCE, since both infras just use an SSH connection.

## How can you test this?

We could deploy it to staging and run a job with tracing enabled.

Did I miss anything?


mjm:
Finally got this working in staging, though all the challenge there was in getting the environment set up correctly, so will merge as soon as CI passes with the updates from master.

PR #514, opened 2018-10-08
Minor README improvements

jsoref:
Minor issues in the Readme.

Random. At the end, I used Grammarly (because this prompt asked me what approach I used, so I picked an approach to fill in here).

## How can you test this?
I didn't.


BanzaiMan:
Looks good. Thank you.

PR #512, opened 2018-10-04
Worker Trace - Opencensus and Stackdriver 

Lyoness:
refs https://github.com/travis-ci/reliability/issues/141

MVP worker trace with exporter and root span. 

## What is the problem that this PR is trying to fix?

Let's  do a first iteration on task/subtask execution & latency on our worker instances ⏲ 

## What approach did you choose and why?
We've already implemented trace in /gcloud-cleanup using opencensus go and the stackdriver exporter, so repeating here. 

## How can you test this?
Running worker locally.  

## What feedback would you like, if any?

We want to be backend agnostic, so:

Ok to put everything in global (rather than provider) config? 

#TODO
Link the associated PRs in terraform and config keychain.

PR #516, opened 2018-10-10
Worker trace: child spans for all steps in ProcessorRun root span

Lyoness:

We merged a stackdriver trace prototype in #512 with a root span that started in `(*Processor).process`
This is a 2nd iteration that adds child spans to all steps added in that func to measure latency in this part of the worker (processor) run. 

ProcessorRun is at the heart of the execution actions in worker.  We wanted to start there. 

## How can you test this?

Tested locally running worker on staging.  The incremental addition of each child span in the steps can be seen here 
webp net-gifmaker

Child spans added have been scoped to [this block of code](https://github.com/travis-ci/worker/blob/master/processor.go#L212-L247) in their individual `step_*` files, including cleanup where applicable.  Are there additional places I should be adding child spans in these files? Omissions?

### DIFF
--- a/step_run_script.go
+++ b/step_run_script.go
@@ -123,7 +123,6 @@ func (s *stepRunScript) Run(state multistep.StateBag) multistep.StepAction {
 }
 
 func (s *stepRunScript) writeLogAndFinishWithState(procCtx, ctx gocontext.Context, logWriter LogWriter, buildJob Job, state FinishState, logMessage string) {
-
 	ctx, span := trace.StartSpan(ctx, "stepWriteLogAndFinishWithState")
 	defer span.End()
 
--- a/step_run_script.go
+++ b/step_run_script.go
@@ -11,6 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/travis-ci/worker/backend"
 	"github.com/travis-ci/worker/context"
+	"go.opencensus.io/trace"
 )
 
 type runScriptReturn struct {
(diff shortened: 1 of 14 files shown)


step_send_received.go
```diff
@@ -25,4 +29,9 @@ func (s *stepSendReceived) Run(state multistep.StateBag) multistep.StepAction {
 	return multistep.ActionContinue
 }
 
-func (s *stepSendReceived) Cleanup(state multistep.StateBag) {}
+func (s *stepSendReceived) Cleanup(state multistep.StateBag) {
+	ctx := state.Get("ctx").(gocontext.Context)
+
+	ctx, span := trace.StartSpan(ctx, "stepSendReceived_cleanup")
+	defer span.End()
```
igorwwwwwwwwwwwwwwwwwwww: This one is not doing anything, we can probably keep it empty.

step_run_script.go
```diff
@@ -119,6 +123,10 @@ func (s *stepRunScript) Run(state multistep.StateBag) multistep.StepAction {
 }
 
 func (s *stepRunScript) writeLogAndFinishWithState(procCtx, ctx gocontext.Context, logWriter LogWriter, buildJob Job, state FinishState, logMessage string) {
+
```
igorwwwwwwwwwwwwwwwwwwww: Extra whitespace, there's one or two more of these.

step_subscribe_cancellation.go
```diff
@@ -25,6 +35,11 @@ func (s *stepSubscribeCancellation) Cleanup(state multistep.StateBag) {
 		return
 	}
 
+	ctx := state.Get("ctx").(gocontext.Context)
+
+	ctx, span := trace.StartSpan(ctx, "SubscribeCancellation_cleanup")
```
igorwwwwwwwwwwwwwwwwwwww: Perhaps we can name these `SubscribeCancellation.Run` and `SubscribeCancellation.Cleanup` respectively? (that would include adjusting everything else to follow that naming convention)

Lyoness:
Hitting this bug: https://github.com/golang/go/issues/26073

Lyoness:
screen shot 2018-10-11 at 8 02 28 am

PR #518, opened 2018-10-12
Opencensus trace span attributes to Processor.Process.Run root span

Lyoness:
Refs https://github.com/travis-ci/reliability/issues/198

Follow up to #512 #516.  Once this is deployed to production, we want to be able to do filter queries in stackdriver based on infra/site, repo, or job_id. 

Since we'll be doing other root spans in https://github.com/travis-ci/reliability/issues/199 that will need different attributes, I didn't want to declare attributes globally for all root spans, kept it to where the root span is declared. 

## How can you test this?

Running worker locally. You'll get these in the stackdriver console.

screen shot 2018-10-12 at 8 25 49 am

As we add more root spans, does it makes sense to have a top-level attribute declaration to omit duplication?  Other?

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -194,7 +194,7 @@ func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
 	span.AddAttributes(
 
 		trace.StringAttribute("app", "worker"),
-		trace.Int64Attribute("JobID", (int64(job_id))),
+		trace.Int64Attribute("job_id", (int64(job_id))),
 		trace.StringAttribute("repo", repo),
 		trace.StringAttribute("infra", infra),
 		trace.StringAttribute("site", site),
--- a/processor.go
+++ b/processor.go
@@ -187,12 +187,11 @@ func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
 	defer span.End()
 
 	span.AddAttributes(
-
 		trace.StringAttribute("app", "worker"),
 		trace.Int64Attribute("job_id", int64(buildJob.Payload().Job.ID)),
-		trace.StringAttribute("repo", string(buildJob.Payload().Repository.Slug)),
-		trace.StringAttribute("infra", string(p.config.ProviderName)),
-		trace.StringAttribute("site", string(p.config.TravisSite)),
+		trace.StringAttribute("repo", buildJob.Payload().Repository.Slug),
+		trace.StringAttribute("infra", p.config.ProviderName),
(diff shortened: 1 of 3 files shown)


processor.go
```diff
@@ -186,8 +186,18 @@ func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
...
 
+	jobID := buildJob.Payload().Job.ID
+	repo := buildJob.Payload().Repository.Slug
+	infra := p.config.ProviderName
+	site := p.config.TravisSite
+
 	span.AddAttributes(
+
 		trace.StringAttribute("app", "worker"),
+		trace.Int64Attribute("JobID", (int64(jobID))),
```
igorwwwwwwwwwwwwwwwwwwww: for consistency, I'd suggest changing the key to `"job_id"`

processor.go
```diff
@@ -186,8 +186,18 @@ func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
...
+	job_id := buildJob.Payload().Job.ID
+	repo := buildJob.Payload().Repository.Slug
+	infra := p.config.ProviderName
+	site := p.config.TravisSite
+
 	span.AddAttributes(
+
 		trace.StringAttribute("app", "worker"),
+		trace.Int64Attribute("JobID", (int64(job_id))),
+		trace.StringAttribute("repo", repo),
```
igorwwwwwwwwwwwwwwwwwwww: these could probably be simplified a bit by inlining the variables from above, wdyt?

igorwwwwwwwwwwwwwwwwwwww: e.g.

```
trace.StringAttribute("repo", buildJob.Payload().Repository.Slug),
```

processor.go
```diff
@@ -187,7 +187,12 @@ func (p *Processor) process(ctx gocontext.Context, buildJob Job) {
 	defer span.End()
 
 	span.AddAttributes(
+
 		trace.StringAttribute("app", "worker"),
+		trace.Int64Attribute("job_id", (int64(buildJob.Payload().Job.ID))),
+		trace.StringAttribute("repo", (string(buildJob.Payload().Repository.Slug))),
+		trace.StringAttribute("infra", (string(p.config.ProviderName))),
+		trace.StringAttribute("site", (string(p.config.TravisSite))),
```
igorwwwwwwwwwwwwwwwwwwww: We can drop the surrounding parentheses around the value here. That would be simplified as:

```
trace.StringAttribute("site", string(p.config.TravisSite)),
```

PR #526, opened 2018-10-17
Expand set of traced functions (ratelimit, backend/gce, GCEapi)

Lyoness:
Refs https://github.com/travis-ci/reliability/issues/199

## What is the problem that this PR is trying to fix?

See the steps and latency of things in our 
* backend/gce infra, including GCEAPI rate limiting
* redis rate limiting

## What approach did you choose and why?
Extension of  #516 

## How can you test this?
Running worker locally and viewing the traces on stackdriver staging project.  An example of an expanded trace is [here](https://console.cloud.google.com/traces/traces?project=travis-staging-1&organizationId=928883094116&q=%2Bapp:worker&tid=dbc795f7d2ec1089f71e12b013ec2c3c&start=1539812529379&end=1539816129379)

## What feedback would you like, if any?

what expanded traces would be useful, what would be noise? 
I tried to trace http requests in image.makeImageRequest in `api_selector.go` , but ran into some problems.

PR #528, opened 2018-10-19
opencensus: fix orphaned spans

igorwwwwwwwwwwwwwwwwwwww:

There's some odd orphaned spans in stackdriver. This happens because the context being passed in is not always an actual child of the `Processor.process` one.

screen shot 2018-10-19 at 17 11 02

I believe some of these were introduced by https://github.com/travis-ci/worker/pull/526.

Some of them are from before the processor runs, during processor startup. For those, we can check that there is a parent span, and only record additional ones if that is the case.

In other cases we are using `procCtx`, a kind of processor-wide context. From what I've been able to tell, we do this to avoid running into issues with context timeouts, where we cannot do critical things like requeues.

My attempt at fixing that replaces `procCtx` with a copy of `ctx` that is made right before the timeout is applied. This way the timeout does not apply for the critical actions, but other properties of the context are still preserved.

I'd definitely like to know if there are any other reasons for the `procCtx`, and whether there is anything that could break by changing it this way.

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -591,8 +591,11 @@ func newGCEProvider(cfg *config.ProviderConfig) (Provider, error) {
 }
 
 func (p *gceProvider) apiRateLimit(ctx gocontext.Context) error {
-	ctx, span := trace.StartSpan(ctx, "apiRateLimit")
-	defer span.End()
+	if trace.FromContext(ctx) != nil {
+		var span *trace.Span
+		ctx, span = trace.StartSpan(ctx, "apiRateLimit")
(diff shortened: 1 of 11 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Confirmed, this is consistent with chirp builds that started around that time:

screen shot 2018-10-19 at 17 42 58

igorwwwwwwwwwwwwwwwwwwww:
And the trace just updated to include all of the parents! 🎉 

screen shot 2018-10-19 at 17 52 04

PR #531, opened 2018-10-23
backend/fake: introduce option to error out all jobs

igorwwwwwwwwwwwwwwwwwwww:
This is in order to test out the behaviour of the system when worker is constantly erroring (requeue storms).

Sample config:

```
export TRAVIS_WORKER_AMQP_URI=...
export TRAVIS_WORKER_BUILD_API_URI=...
export TRAVIS_WORKER_QUEUE_NAME=builds.standalone
export TRAVIS_WORKER_AMQP_CONSUMER_PRIORITY=1
export TRAVIS_WORKER_POOL_SIZE=50

export TRAVIS_WORKER_PROVIDER_NAME=fake
export TRAVIS_WORKER_FAKE_ERROR=true
```

refs https://github.com/travis-ci/reliability/issues/208

### DIFF
--- a/backend/fake.go
+++ b/backend/fake.go
@@ -2,6 +2,7 @@ package backend
 
 import (
 	"context"
+	"errors"
 	"io"
 	"time"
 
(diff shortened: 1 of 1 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Seeing some weird behaviour here. After a couple requeues, the job gets stuck where worker thinks it's finished, but the main database thinks it's still running. :-/

I wonder if this is a race condition in hub, perhaps what https://github.com/travis-pro/team-teal/issues/631 was trying to address?

PR #533, opened 2018-10-24
backend/gce: annotate trace with ssh attempt error

igorwwwwwwwwwwwwwwwwwwww:
This should allow us to tell what the SSH dial attempts are failing with during `boot_poll_ssh`.

It would be interesting to see if it is networking in general, or something SSH specific.

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -1453,7 +1453,15 @@ func (i *gceInstance) sshConnection(ctx gocontext.Context) (remote.Remoter, erro
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Innnnnnnteresting, from staging:
```
couldn't connect to SSH server: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

igorwwwwwwwwwwwwwwwwwwww:
Oh, and before that we have two slower calls of:

```
couldn't connect to SSH server: dial tcp 10.20.0.8:22: i/o timeout
```

With a 5 second timeout each.

PR #537, opened 2018-10-30
introduce http_proxy variable in docker container

igorwwwwwwwwwwwwwwwwwwww:

This is to support http proxies in enterprise, in collaboration with @schultyy.

Setting an environment variable on the docker container. In the long-term, it would probably be better to put this kind of logic into `travis-build` (especially to make this work across infras), but from what I've heard it is currently easier to roll out a newer version of worker on enterprise than travis-build.

So that's what we're going with for now.

## How can you test this?

I'm testing it with a local worker. We can also test it with our enterprise test installation.

It can be tested by setting the environment variable:

```
export TRAVIS_WORKER_DOCKER_HTTP_PROXY=http://proxy:80880
```

That should make the `http_proxy` and `HTTP_PROXY` env vars set in the container.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ## [Unreleased]
 
 ### Added
+- backend/docker: support for setting `HTTP_PROXY` variable in jobs
 
 ### Changed
 -trace:
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -99,6 +99,7 @@ type dockerProvider struct {
 	tmpFs           map[string]string
 	imageSelector   image.Selector
 	containerLabels map[string]string
+	httpProxy       string
 
 	cpuSetsMutex sync.Mutex
 	cpuSets      []bool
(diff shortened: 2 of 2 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Getting some docker version issues with my local docker install:

```
time="2018-10-30T14:11:29+01:00" level=error msg="couldn't run script, attempting requeue" completed=false err="\"env\" requires API version 1.25, but the Docker daemon API version is 1.24" instance_id=08ce1a9 job_id=730669 job_path=igorwwwwwwwwwwwwwwwwwwww/hello-world/jobs/730669 pid=85514 processor=407e24e0-5d35-4d63-99c8-aa7d866ab18a@85514.gopher.local repository=igorwwwwwwwwwwwwwwwwwwww/hello-world self=step_run_script
```

igorwwwwwwwwwwwwwwwwwwww:
We have a worker image to test with: `travisci/worker:v4.5.2-12-ga556057`.

igorwwwwwwwwwwwwwwwwwwww:
Looks like we need to set the API version the client uses. There's a config var for that:

```
export TRAVIS_WORKER_DOCKER_API_VERSION=1.37
```

igorwwwwwwwwwwwwwwwwwwww:
Confirmed working!

> Timeout waiting for network availability.
> ...
> $ echo $http_proxy
> http://proxy:80880

schultyy:
Tested this on a test installation for Enterprise and everything works out of the box. ✔️

klausmeyer-fidor:
Thanks guys 🍺

PR #540, opened 2018-11-01
backend:gce opencensus trace- add time.Sleep calls to processor.Run root span

Lyoness:
Refs https://github.com/travis-ci/reliability/issues/199

Even though we have sleep times as constants during a job run, we don't have any indication of how many times those sleeps are called as a result of retries.  This adds them to worker trace to give us that data. 

Passing in context to `time.Sleep` calls in the same manner as https://github.com/travis-ci/worker/pull/528, so that we don't orphan any outside of the Processor.Run 

## How can you test this?

Staging worker with time.Sleep spans added [here](https://console.cloud.google.com/traces/traces?project=travis-staging-1&tid=ba1e9379f01be950fb19af218ee8c549&start=1541075763880&end=1541079363880&q=ProcessorRun) 

screen shot 2018-11-01 at 9 46 56 am

I'm not entirely sure if this is the way I should be doing this, I simply followed patterns and this is the only thing that compiled. 😅

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -644,6 +644,12 @@ func (p *gceProvider) apiRateLimit(ctx gocontext.Context) error {
 
 		// Sleep for up to 1 second
 		time.Sleep(time.Millisecond * time.Duration(mathrand.Intn(1000)))
+
+		if trace.FromContext(ctx) != nil {
+			var span *trace.Span
+			ctx, span = trace.StartSpan(ctx, "GCE.timeSleep.apiRateLimit")
+			defer span.End()
+		}
 	}
 }
 
@@ -1062,6 +1068,12 @@ func (p *gceProvider) stepWaitForInstanceIP(c *gceStartContext) multistep.StepAc
 	})
 
 	time.Sleep(p.bootPrePollSleep)
+
+	if trace.FromContext(ctx) != nil {
+		var span *trace.Span
+		ctx, span = trace.StartSpan(ctx, "GCE.timeSleep.WaitForInstanceIP")
+		defer span.End()
+	}
 
 	zoneOpCall := p.client.ZoneOperations.Get(p.projectID, c.zoneName, c.instanceInsertOpName).Context(c.ctx)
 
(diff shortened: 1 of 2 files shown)


Lyoness:
Better! This is actually a cool way to visualize `defer` function calls in a stack trace btw
screen shot 2018-11-02 at 7 59 32 am

backend/gce.go
```diff
@@ -643,7 +643,11 @@ func (p *gceProvider) apiRateLimit(ctx gocontext.Context) error {
 		}
 
 		// Sleep for up to 1 second
+		var span *trace.Span
+		ctx, span = trace.StartSpan(ctx, "GCE.timeSleep.apiRateLimit")
 		time.Sleep(time.Millisecond * time.Duration(mathrand.Intn(1000)))
+		span.End()
```
igorwwwwwwwwwwwwwwwwwwww: This one I think we need to do some `if trace.FromContext(ctx) != nil {` messiness to avoid orphaned spans.

The issue is that `apiRateLimit` is sometime called from outside the ProcessorRun root span, e.g. during setup. So we need to make sure to only add a child span if there actually is a parent.

The other trickiness here is that we need to make the `span.End()` conditional too, without using `defer`.

backend/gce.go
```diff
@@ -644,10 +644,15 @@ func (p *gceProvider) apiRateLimit(ctx gocontext.Context) error {
 
 		// Sleep for up to 1 second
 		var span *trace.Span
-		ctx, span = trace.StartSpan(ctx, "GCE.timeSleep.apiRateLimit")
+		if trace.FromContext(ctx) != nil {
+			ctx, span = trace.StartSpan(ctx, "GCE.timeSleep.apiRateLimit")
+		}
+
 		time.Sleep(time.Millisecond * time.Duration(mathrand.Intn(1000)))
```
Lyoness: I kept this `time.Sleep` outside of any conditionals, because we still need it to run independent of whether a context is present or not for tracing.

Lyoness: I kept this time.Sleep outside of any conditionals, because we still need it to run independent of whether a context is present or not for tracing.

backend/gce.go
```diff
@@ -644,10 +644,15 @@ func (p *gceProvider) apiRateLimit(ctx gocontext.Context) error {
...
-		span.End()
 
+		if trace.StartSpan != nil {
```
igorwwwwwwwwwwwwwwwwwwww: you can probably test for `if span != nil` here :)

PR #541, opened 2018-11-02
backend/gce: track metric for rate of calls to gce api

igorwwwwwwwwwwwwwwwwwwww:
The goal is to get a better idea of how many calls we are actually sending to the API, since the GCE quotas page only gives us 5 minute averages, and we'll want to know more about the peaks.


igorwwwwwwwwwwwwwwwwwwww:
Giving this a test run on a production worker.

igorwwwwwwwwwwwwwwwwwwww:
Seems to be working! Going to go ahead and merge this. :)

PR #543, opened 2018-11-05
Do not override context in existing child

Lyoness:
passing context through a var declaration with the same name as the parent in a loop results in grandchildren-of-children without being tied to their parents.

This weird rendering of a loop child span [here](https://console.cloud.google.com/traces/traces?organizationId=928883094116&project=eco-emissary-99515&tid=0257f175c9cf5650c303e8fc4af246fd&q=%2Bapp:worker&start=1541430390845&end=1541433990845)

use a blank identifier to ignore a new context declaration. 

## How can you test this?
ran a worker in staging, this fixes the span rendering. ✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Removed
 
 ### Fixed
+- trace:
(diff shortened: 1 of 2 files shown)


Lyoness:
✅ 
screen shot 2018-11-05 at 1 06 39 pm

PR #548, opened 2018-11-07
trace: trace CLI.Setup and avoid orphaned spans during worker boot

igorwwwwwwwwwwwwwwwwwwww:
We've been seeing some orphaned spans during worker boot. This ties them together:

screen shot 2018-11-07 at 16 57 27

The patch also removes an instance deletion that happens in `gce.go`. That really should be done via `StartInstance.Cleanup`. If there are any cases in which `StartInstance.Cleanup` does not properly delete the instance, I'd love to understand those. But from what I can tell, it should always take care of things.


igorwwwwwwwwwwwwwwwwwwww:
Rolled out to staging.

PR #554, opened 2018-11-15
remove cloudbrain backend (so long)

igorwwwwwwwwwwwwwwwwwwww:
And thanks for all the 🐟


igorwwwwwwwwwwwwwwwwwwww:
Note: this means our next release should be v5.0.0 since this is Technically a BC break.

PR #558, opened 2018-11-21
Adding zone information when using warmed instances

ryndaniels:
VMs from warmer often don't get deleted because the worker ends up looking in the wrong zone when trying to delete them.

Passing back the zone information from warmer and then using that in the gce worker backend for the deletion.

## How can you test this?
Turn worker back on, see if it's able to delete instances this time

How GOod is my Go code?


igorwwwwwwwwwwwwwwwwwwww:
Oh this also needs an entry in CHANGELOG.md.

PR #560, opened 2018-11-22
ratelimit: dynamic configuration of `max_calls` and `duration` via redis

igorwwwwwwwwwwwwwwwwwwww:
In order to be able to adjust our rate limiting on the fly it would be good to be able to dynamically configure it through redis.

This allows the values to be changed without deploying or restarting worker. This will allow us to slowly approach our API quotas.

The main risk is increased traffic to redis. I think redis should be able to handle a couple hundred requests per second easily though. So I'm not too worried. In the worst case we can cache the config value in memory though and update it every couple seconds (maybe using something like [go-cache](https://github.com/patrickmn/go-cache)).

To enable in worker (terraform-config):

```
TRAVIS_WORKER_GCE_RATE_LIMIT_DYNAMIC_CONFIG=true
```

To set the values in redis:

```
trvs redis-cli travis-production REDIS_URL mset worker-rate-limit-gce-api-1:gce-api:max_calls 3000 worker-rate-limit-gce-api-1:gce-api:duration 100s
```

That should make the rate-limiter perform 3000 calls per 100 seconds across all workers in `production-1`.

refs https://github.com/travis-ci/reliability/issues/222

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ## [Unreleased]
 
 ### Added
+- ratelimit: dynamic configuration of `max_calls` and `duration` via redis
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


igorwwwwwwwwwwwwwwwwwwww:
I guess this is pretty close to a circuit breaker. :)

PR #565, opened 2018-11-29
Add Redis-based concurrent job limiter

mjm:

On MacStadium, we have a fixed amount of machines, so we need a fixed limit on how many concurrent jobs we run. We've been managing this with pool sizes, but that causes issues when doing new rollouts on Kubernetes, as we can't wait for graceful shutdown before bringing up new workers.

See travis-ci/kubernetes-config#5 for more details.

Use Redis to track as jobs are started and stopped across the different worker instances. When starting a job, total all the job counts across each worker, and compare it to a total allowed number of jobs defined in a file on disk (will come from a ConfigMap when deployed in Kubernetes). Only proceed with running a job if there is room for more jobs. If not, sleep for 10-20 seconds and try again.

## How can you test this?

Deploy it to macstadium-staging and experiment with running concurrent jobs.

Any. One known limitation with this implementation is that (AFAICT) this does the sleep loop after the job has already been claimed off the AMQP queue. For the most part that's fine, but if we try to do a graceful shutdown of a worker in this state, it will wait to run the job before the processor can finish. If we were able to wait before pulling the job off the queue, then the job would just go to a different worker. I don't know if this is possible or how to do it though.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -58,6 +58,7 @@ type CLI struct {
 
 	Config                  *config.Config
 	BuildScriptGenerator    BuildScriptGenerator
+	ConcurrentJobLimiter    ConcurrentJobLimiter
 	BuildTracePersister     BuildTracePersister
 	BackendProvider         backend.Provider
 	ProcessorPool           *ProcessorPool
@@ -161,6 +162,8 @@ func (i *CLI) Setup() (bool, error) {
 
 	i.BuildTracePersister = persister
 
+	i.setupConcurrentJobLimiter()
+
 	if i.Config.TravisSite != "" {
 		i.Config.ProviderConfig.Set("TRAVIS_SITE", i.Config.TravisSite)
 	}
@@ -187,7 +190,7 @@ func (i *CLI) Setup() (bool, error) {
 		Config:   i.Config,
 	}
 
-	pool := NewProcessorPool(ppc, i.BackendProvider, i.BuildScriptGenerator, i.BuildTracePersister, i.CancellationBroadcaster)
+	pool := NewProcessorPool(ppc, i.BackendProvider, i.BuildScriptGenerator, i.ConcurrentJobLimiter, i.BuildTracePersister, i.CancellationBroadcaster)
 
 	logger.WithField("pool", pool).Debug("built")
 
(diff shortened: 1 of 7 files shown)


mjm:
The Kubernetes config changes to use this are here: https://github.com/travis-ci/kubernetes-config/pull/27. I've tested them in staging and they seem to behave as expected.

When deploy this for production, it will only affect the workers that are running in Kubernetes, and we will set the total job count accordingly. That should mitigate any potential issues that could arise from this.

soulshake:
> We've been managing this with pool sizes, but that causes issues when doing new rollouts on Kubernetes, as we can't wait for graceful shutdown before bringing up new workers.

Can you elaborate on this? Why would we have to wait for shutdown before bringing up new workers when using pool sizes? Isn't there a way to make Kubernetes create the new worker before removing the old one?

mjm:
@soulshake I sure can!

There is a way, and we use it. The issue here is pretty subtle, and it only really matters in environments like MacStadium, where we have a fixed amount of computing capacity to work with. We've done the math and know exactly how many jobs we are willing to run at any one time to not oversaturate the hardware.

Pre-Kubernetes, we managed this by configuring worker's pool size such that all of the pools across all workers sum to the total number of jobs we want to run. When we deployed new worker code/config, a human would have to iterate through the workers, sending a single worker at a time a graceful termination signal and waiting for it to drain its pool and restart. Once it had restarted and was back at full capacity, the human could gracefully restart the next worker, and so on for a few hours until it was all done.

If we take a similar approach to managing the pool in Kubernetes, though, we have an issue when we deploy new versions: once Kubernetes sends the signal to terminate a worker, it doesn't wait for the pod to actually terminate before bringing up a new one to replace it. It thinks its work is done, and doesn't expect terminating a pod to take 2 hours or something like that. So it moves on to the next pod, tells it to terminate, and starts a new one to replace it.

So if we started with 4 workers in the deployment, after we deploy the new version, we will have 8 workers running: 4 that are working on gracefully draining their pool of jobs, and 4 brand new ones with their full pool of processors. If we got a surge of jobs during this time, in the worst case scenario, we could run twice as many concurrent jobs as we want to!

I tried a bunch of things, but as far as I can tell, there isn't a native Kubernetes way to get the rollout to be staggered in the way we want. So instead, we've done this to give worker the ability to manage its concurrency of the entire pool on its own. It lets us do deployments the normal Kubernetes way, and makes the number of workers and their pool sizes independent of how many jobs we actually can run, which I think gives us some more flexibility in how we deploy.

mjm:
This was superseded by https://github.com/travis-ci/worker-operator

PR #566, opened 2018-12-06
Make worker pool size manageable by software

mjm:

To run worker correctly in Kubernetes, we need to be able to control rollouts of new instances of worker such that the pool size doesn't exceed our computing capacity for running CI jobs. We're writing a Kubernetes controller for that, but to do its job, worker needs to have API for controlling the pool size. Some of that was present in the existing HTTP API, but it was missing some important things.

* Refactored HTTP API to use a Gorilla mux and have its own struct with the API handlers
* Changed the API URLs, methods, and format to be more resource-oriented and use JSON for easier use by other software
* Added API to ProcessorPool to set the size of the processor pool, instead of only being able to increment and decrement
* Firmed up the contract around `ProcessorPool.Size()`: it now includes processors that are still waiting to gracefully shutdown. Also added the old concept (which excludes processors that have been told to shutdown) back under a new name: `ExpectedSize()` (as it is expected to be the eventual size of the pool)
* Changed behavior when pool size is configured to be 0: now worker will wait for the first processor to start before falling through to the wait group. This lets the pool size be entirely managed by an outside service. In this mode, worker doesn't know what its pool size should be, and is waiting for another process to tell it how many processors it should start.

## How can you test this?

I'm testing this with an extra deployment in macstadium-staging, making manual requests to the HTTP API to make adjustments to the pool size.

Any.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -100,19 +100,6 @@ func (i *CLI) Setup() (bool, error) {
 
 	i.Config = config.FromCLIContext(i.c)
 
-	if i.c.String("remote-controller-addr") != "" {
-		if i.c.String("remote-controller-auth") != "" {
-			i.setupRemoteController()
-		} else {
-			i.logger.Info("skipping remote controller setup without remote-controller-auth set")
-		}
-		go func() {
-			httpAddr := i.c.String("remote-controller-addr")
-			i.logger.Info("listening at ", httpAddr)
-			http.ListenAndServe(httpAddr, nil)
-		}()
-	}
-
 	if i.c.Bool("echo-config") {
 		config.WriteEnvConfig(i.Config, os.Stdout)
 		return false, nil
@@ -182,6 +169,19 @@ func (i *CLI) Setup() (bool, error) {
 	logger.WithField("pool", pool).Debug("built")
 
 	i.ProcessorPool = pool
+
+	if i.c.String("remote-controller-addr") != "" {
+		if i.c.String("remote-controller-auth") != "" {
+			i.setupRemoteController()
+		} else {
+			i.logger.Info("skipping remote controller setup without remote-controller-auth set")
+		}
+		go func() {
+			httpAddr := i.c.String("remote-controller-addr")
+			i.logger.Info("listening at ", httpAddr)
+			http.ListenAndServe(httpAddr, nil)
+		}()
+	}
 
 	err = i.setupJobQueueAndCanceller()
 	if err != nil {
--- a/cli.go
+++ b/cli.go
@@ -498,8 +498,32 @@ func (i *CLI) heartbeatCheck(heartbeatURL, heartbeatAuthToken string) error {
 }
 
 func (i *CLI) setupHTTPAPI() {
-	apiHandler := &APIHandler{i: i}
+	i.logger.Info("setting up HTTP API")
+	apiHandler := &APIHandler{
+		pool:       i.ProcessorPool,
+		auth:       i.c.String("http-api-auth"),
+		workerInfo: i.workerInfo,
+		cancel:     i.cancel,
+	}
 	apiHandler.Setup()
+}
+
+func (i *CLI) workerInfo() workerInfo {
+	info := workerInfo{
+		Version:          VersionString,
+		Revision:         RevisionString,
+		Generated:        GeneratedString,
+		Uptime:           time.Since(i.bootTime).String(),
+		PoolSize:         i.ProcessorPool.Size(),
+		ExpectedPoolSize: i.ProcessorPool.ExpectedSize(),
+		TotalProcessed:   i.ProcessorPool.TotalProcessed(),
+	}
+
+	i.ProcessorPool.Each(func(_ int, p *Processor) {
+		info.Processors = append(info.Processors, p.processorInfo())
+	})
+
+	return info
 }
 
 func (i *CLI) signalHandler() {
--- a/cli.go
+++ b/cli.go
@@ -1,7 +1,6 @@
 package worker
 
 import (
-	"crypto/subtle"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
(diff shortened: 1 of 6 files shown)


cli.go
```diff
@@ -117,7 +116,7 @@ func (i *CLI) Setup() (bool, error) {
 			if httpPort == "" {
 				httpPort = i.c.String("pprof-port")
 			}
-			httpAddr := fmt.Sprintf("localhost:%s", httpPort)
+			httpAddr := fmt.Sprintf("0.0.0.0:%s", httpPort)
```
meatballhat: I would prefer a change to an explicit `"http-addr"` option, defaulting to `127.0.0.1:1627` (`'travis-ci/worker'.scan(/./).map(&:ord).sum`). Said another way, I don't think that the default should be for the API to serve on all interfaces.

http_api.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+// APIHandler handles requests to worker's HTTP API.
+type APIHandler struct {
+	i *CLI
+}
+
+// Setup installs the HTTP routes that will handle requests to the HTTP API.
+func (api *APIHandler) Setup() {
+	api.i.logger.Info("setting up HTTP API")
```
meatballhat: This is the only place where `api.i.logger` is accessed, so maybe log this up in `cli.go` when running `Setup`?

http_api.go
```diff
@@ -0,0 +1,194 @@
...
+
+		username, password, ok := req.BasicAuth()
+		if !ok {
+			w.Header().Set("WWW-Authenticate", "Basic realm=\"travis-ci/worker\"")
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		authBytes := []byte(fmt.Sprintf("%s:%s", username, password))
+		if subtle.ConstantTimeCompare(authBytes, []byte(api.i.c.String("http-api-auth"))) != 1 {
```
meatballhat: This is the only place where `api.i.c` is accessed, so I think I would prefer it be injected as a string value instead of reaching through `.i.c.String`.

http_api.go
```diff
@@ -0,0 +1,194 @@
...
+	"strings"
+	"time"
+
+	"github.com/gorilla/mux"
+)
+
+// APIHandler handles requests to worker's HTTP API.
+type APIHandler struct {
+	i *CLI
+}
```
meatballhat: There are some other related comments, but I think I'm leaning toward this struct having access/construction more like:

``` go
type APIHandler struct {
  processorPool *ProcessorPool
  auth string
  info func() *workerInfo
  cancel func()
}
```

http_api.go
```diff
@@ -0,0 +1,194 @@
...
+	Generated        string `json:"generated"`
+	Uptime           string `json:"uptime"`
+	PoolSize         int    `json:"poolSize"`
+	ExpectedPoolSize int    `json:"expectedPoolSize"`
+	TotalProcessed   int    `json:"totalProcessed"`
+
+	Processors []processorInfo `json:"processors"`
+}
+
+func (i *CLI) workerInfo() workerInfo {
```
meatballhat: I would like these methods to live alongside their respective struct definitions :heart:

cli.go
mjm: Okay, that seems reasonable.

http_api.go
mjm: Okay, I can get behind that. The last two function strike me as a little strange to bring in this way, but I'm not sure. Maybe we can reconsider later if we end up needing more of `CLI`'s methods in the future.

cli.go
mjm: There may not be need for a default here: right now, passing a port is required to even enable this API in the first place. With this change, you would have to specify the whole address including the port, to enable the API. So we don't need a default address/port unless we start enabling the HTTP API by default. Should we?

mjm: I handled this by removing the `pprof-port` and `http-api-port` config defs and replaced them both with `http-api-addr`. It really simplifies the logic here, which definitely confused me the first time I was trying to figure out what was needed to turn on the HTTP API.

cli.go
```diff
@@ -500,100 +490,31 @@ func (i *CLI) heartbeatCheck(heartbeatURL, heartbeatAuthToken string) error {
...
-		w.Header().Set("WWW-Authenticate", "Basic realm=\"travis-ci/worker\"")
-		w.WriteHeader(http.StatusUnauthorized)
-		return
+	apiHandler := &APIHandler{
+		pool:       i.ProcessorPool,
+		auth:       i.c.String("http-api-auth"),
+		workerInfo: i.workerInfo,
+		cancel:     i.cancel,
 	}
+	apiHandler.Setup()
```
meatballhat: I'm sorry for not catching it before, but I realize this `apiHandler` is not retained as a reference, which maybe doesn't matter (???) but it's definitely confusing for me when looking at it now. Would it be handy to keep a reference, a la `i.apiHandler`?

http_api.go
```diff
@@ -0,0 +1,168 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/gorilla/mux"
+)
+
+// APIHandler handles requests to worker's HTTP API.
+type APIHandler struct {
```
meatballhat: This is definitely a nitpick, but I think that this should be changed to `HTTPAPI` or `HttpApi` or the file name should be `api_handler.go` (or something else?)

cli.go
mjm: It should be fine, because the handler methods should keep a reference to the struct, and the HTTP serve mux is holding a reference to those handlers.

We could store it on the `CLI` if you think that's less confusing to a new reader of the code, but I don't think we'll use the reference anywhere later.

meatballhat: Sorry sorry, I should have been clear that I wasn't worried about GC, but human readers. Given there's no need for a reference, WDYT about skipping an lvar entirely like `(&Thing{}).DoStuff()`?

http_api.go
mjm: Yeah, I think I almost called it `HTTPAPI` and balked because of the EXCESSIVE CAPS, but it's probably a better name for it even so.

meatballhat: Maybe `WebApp`/`web_app.go`?

mjm:
Oh no:

```
2018/12/07 13:40:53 http: panic serving 10.244.0.0:34438: runtime error: invalid memory address or nil pointer dereference
goroutine 43 [running]:
net/http.(*conn).serve.func1(0xc00073a640)
	/usr/local/go/src/net/http/server.go:1746 +0xd0
panic(0x1006a20, 0x1d21d60)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/travis-ci/worker.(*ProcessorPool).SetSize(0x0, 0x5)
	/go/src/github.com/travis-ci/worker/processor_pool.go:103 +0x26
github.com/travis-ci/worker.(*RemoteController).UpdateWorkerInfo(0xc000634150, 0x1373ca0, 0xc00051c380, 0xc00013c800)
	/go/src/github.com/travis-ci/worker/remote_controller.go:102 +0x142
github.com/travis-ci/worker.(*RemoteController).UpdateWorkerInfo-fm(0x1373ca0, 0xc00051c380, 0xc00013c800)
	/go/src/github.com/travis-ci/worker/remote_controller.go:28 +0x48
net/http.HandlerFunc.ServeHTTP(0xc00023a8a0, 0x1373ca0, 0xc00051c380, 0xc00013c800)
	/usr/local/go/src/net/http/server.go:1964 +0x44
github.com/travis-ci/worker.(*RemoteController).CheckAuth.func1(0x1373ca0, 0xc00051c380, 0xc00013c800)
	/go/src/github.com/travis-ci/worker/remote_controller.go:64 +0x282
net/http.HandlerFunc.ServeHTTP(0xc000228560, 0x1373ca0, 0xc00051c380, 0xc00013c800)
	/usr/local/go/src/net/http/server.go:1964 +0x44
github.com/travis-ci/worker/vendor/github.com/gorilla/mux.(*Router).ServeHTTP(0xc000644070, 0x1373ca0, 0xc00051c380, 0xc00013c800)
	/go/src/github.com/travis-ci/worker/vendor/github.com/gorilla/mux/mux.go:162 +0xf1
net/http.(*ServeMux).ServeHTTP(0x1d525e0, 0x1373ca0, 0xc00051c380, 0xc00013c600)
	/usr/local/go/src/net/http/server.go:2361 +0x127
net/http.serverHandler.ServeHTTP(0xc0005d8680, 0x1373ca0, 0xc00051c380, 0xc00013c600)
	/usr/local/go/src/net/http/server.go:2741 +0xab
net/http.(*conn).serve(0xc00073a640, 0x1374de0, 0xc000631840)
	/usr/local/go/src/net/http/server.go:1847 +0x646
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:2851 +0x2f5
```

Something in the changes from the feedback introduced a regression.

mjm:
All sorted out, appears to be working as expected!

PR #570, opened 2018-12-13
Add ready check endpoint for remote controller

mjm:

Sometimes the worker operator makes requests to the HTTP API before the processor pool has started. This causes a crash because some properties of the ProcessorPool are not set until it's running.

Add a ready check endpoint to the API which checks the pool for its readiness. We'll define a readiness probe for worker pods in Kubernetes so that the pod will not be considered ready until this check passes. It makes sense that the worker pod as a whole wouldn't be ready until the processor pool is up and running.

## How can you test this?

Test in staging with the worker-operator.

Any.

### DIFF
--- a/processor_pool.go
+++ b/processor_pool.go
@@ -84,6 +84,12 @@ func (p *ProcessorPool) Each(f func(int, *Processor)) {
 	}
 }
 
+// Ready returns true if the processor pool is running as expected.
+// Returns false if the processor pool has not been started yet.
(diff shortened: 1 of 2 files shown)


mjm:
Hmm...this is always returning Not Ready when I use this in staging. It's possible that one of the properties I'm checking is not always present even when operating normally.

mjm:
Cool, that fixed it.

PR #572, opened 2018-12-13
Retry instance insert in alternate zone on error

meatballhat:
if instance is not "zone-pinned" such as with vm config override(s).

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -1072,7 +1072,7 @@ func (p *gceProvider) stepInsertInstance(c *gceStartContext) multistep.StepActio
 		c.instance = inst
 		c.instanceInsertOpName = op.Name
 		return nil
-	}, backoff.WithContext(b, ctx).Backoff)
+	}, backoff.WithContext(b, ctx))
 
 	if err != nil {
 		c.progresser.Progress(&ProgressEntry{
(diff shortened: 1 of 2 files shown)


backend/gce.go
```diff
@@ -1029,8 +1049,31 @@ func (p *gceProvider) stepInsertInstance(c *gceStartContext) multistep.StepActio
...
-	op, err := p.client.Instances.Insert(p.projectID, c.zoneName, inst).Context(c.ctx).Do()
+	b := backoff.NewExponentialBackOff()
+	b.InitialInterval = 1 * time.Second
+	b.MaxElapsedTime = 1 * time.Minute
+
+	err = backoff.Retry(func() error {
+		p.apiRateLimit(c.ctx)
+		op, err := p.client.Instances.Insert(p.projectID, c.zoneName, inst).Context(c.ctx).Do()
+		if err != nil {
+			if !c.zonePinned {
```
mjm: I'm curious: is it most likely that an error here is because of an issue with the zone? Do we expect that switching to another will usually fix our issue? Is there anything we could look for in the error to determine if that makes sense?

meatballhat: It's one of the top 3 errors, and I couldn't think of a reason why switching zones/retrying would be a bad idea. I'm hesitant to get more specific based on what's in the error given we don't have error types on which we can rely.

mjm: Ok, that makes sense! 👍

PR #573, opened 2018-12-13
Prepare for v6.0.0 release

mjm:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,28 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 
 ### Changed
-- ratelimit: trace redis connection pool checkout
 
 ### Deprecated
 
 ### Removed
 
 ### Fixed
+
+## [6.0.0] - 2018-12-13
+
+### Added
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,14 +7,28 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
...
+## [6.0.0] - 2018-12-13
+
+### Added
+- rc: remote controller HTTP API
+- processor: set size of processor pool
+
+### Changed
+- ratelimit: trace redis connection pool checkout
+
+### Removed
```
Lyoness: maybe be explicit that this is a breaking change with the old http api? Or add a new section header that says it’s a breaking change?  I'm not sure, but there might be enterprise customers that choose the http queue type.

mjm: I don't think this is the same as the HTTP queue API. I'll confirm, but I think those are unrelated to each other.

mjm: Ok, I checked. The HTTP queue type doesn't expose an HTTP server from worker, it polls job board for new jobs. So this should not affect that.

PR #581, opened 2019-02-11
Fix docs links for timeouted job

iriberri:
A broken link to our docs, where the link doesn't send users to the right anchor due to case-sensitiveness :)

## How can you test this?
See:

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received (working)
vs
https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received


meatballhat:
TODO: don't have a deep link in the code :sweat_smile:

PR #583, opened 2019-04-01
randomize initial zone selection

joshk:

Workers, when deployed, are bound to a primary zone. Whenever a job (and VM) is started, the primary zone is used first, then an alternate zone is selected.

The core problem is when we start hitting zone exhaustion errors for a zone. Since there are pools of workers per zone, and each worker will try its primary zone first, which just puts more pressure on the api, raising the risk of api rate limit issues.

This is a first step towards a fix, but not yet the full fix.

This changes the concept of a primary zone, and instead has a random zone picked each try. A zone can still be defined in the config, but this zone will be used each and every time.

The follow up work to this PR is to block a zone for 10mins across all workers (possibly using Redis), which will help reduce api usage, and a better self healing worker setup.

## How can you test this?

I've tested this locally by connecting it to staging, and it worked a treat.

This is for general discussion

I also need help understanding how to fix the test failures

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -312,6 +312,10 @@ func newGCEProvider(cfg *config.ProviderConfig) (Provider, error) {
 		return nil, err
 	}
 
+	if !cfg.IsSet("ACCOUNT_JSON") {
+		return nil, fmt.Errorf("missing ACCOUNT_JSON")
+	}
+
 	projectID := cfg.Get("PROJECT_ID")
 	if metadata.OnGCE() {
 		projectID, err = metadata.ProjectID()
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -48,7 +48,6 @@ import (
 )
 
 const (
-	defaultGCEZone               = "us-central1-a"
 	defaultGCEMachineType        = "n1-standard-2"
 	defaultGCEPremiumMachineType = "n1-standard-4"
 	defaultGCENetwork            = "default"
@@ -59,7 +58,6 @@ const (
 	defaultGCEStopPollSleep      = 3 * time.Second
 	defaultGCEStopPrePollSleep   = 15 * time.Second
 	defaultGCESubnet             = "default"
-	defaultGCERegion             = "us-central1"
 	defaultGCEUploadRetries      = uint64(120)
 	defaultGCEUploadRetrySleep   = 1 * time.Second
 	defaultGCEImageSelectorType  = "env"
@@ -95,7 +93,7 @@ var (
 		"NETWORK":                fmt.Sprintf("network name (default %q)", defaultGCENetwork),
 		"PREEMPTIBLE":            "boot job instances with preemptible flag enabled (default false)",
 		"PREMIUM_MACHINE_TYPE":   fmt.Sprintf("premium machine type (default %q)", defaultGCEPremiumMachineType),
-		"PROJECT_ID":             "[REQUIRED] GCE project id",
+		"PROJECT_ID":             "[REQUIRED] GCE project id (will try to auto detect it if not set)",
 		"PUBLIC_IP":              "boot job instances with a public ip, disable this for NAT (default true)",
 		"PUBLIC_IP_CONNECT":      "connect to the public ip of the instance instead of the internal, only takes effect if PUBLIC_IP is true (default true)",
 		"IMAGE_PROJECT_ID":       "GCE project id to use for images, will use PROJECT_ID if not specified",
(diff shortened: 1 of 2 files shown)


soulshake:
> The core problem is when we start hitting zone exhaustion errors for a zone. 

Which zone exhaustion errors are you referring to? IIUC, most resource quotas (as well as API quotas) are per project and/or per region rather than per zone. (See [GCE Quotas page](https://console.cloud.google.com/iam-admin/quotas))

Example from recent outage:
```
Error
QUOTA_EXCEEDED: Quota 'SSD_TOTAL_GB' exceeded. Limit: 800000.0 in region us-central1. 
```


> Since there are pools of workers per zone, and each worker will try its primary zone first, which just puts more pressure on the api, raising the risk of api rate limit issues.

I think this is only the case when a zone is specified in the worker config. In our case it looks like instances are pretty well distributed across zones already, no?
image

I wonder if it would make more sense to have each worker create job instances in its own zone, if possible, since those are already automatically distributed by the [managed instance group](https://console.cloud.google.com/compute/instanceGroups/list).

Also, not all resource types (in particular [GPUs](https://cloud.google.com/compute/docs/gpus/)) are available in all zones. So I'm not sure it makes sense to make zone pinning an all-or-nothing thing, because I believe that will cause problems when GPUs (and perhaps some specific CPUs) are specified.

soulshake:
Edit: I take it all back, I see there were plenty of these errors during the last outage:

```
Mar 29 23:32:53 production-2-worker-org-gce-4mr9 travis-worker-wrapper: 
time="2019-03-30T04:32:53Z" level=error msg="couldn't start instance, attempting 
requeue" err="code=ZONE_RESOURCE_POOL_EXHAUSTED location= 
message=The zone 'projects/travis-ci-prod-2/zones/us-central1-c' does not have 
enough resources available to fulfill the request.  Try a different zone, or try again later." 
job_id=123456798 job_path=xyz/xyz/jobs/123456798 pid=1 
processor=ed8d48ea-5209-4f2b-b595-bfda4c06ce13@1.production-2-worker-org-gce-4mr9 
repository=xyz/xyz self=step_start_instance start_timeout=8m0s uuid=84c6f99e-f021-4b13-baf1-ba101c22e3ab
```

joshk:
Thanks for the feedback AJ.

I'm just about to hit the hay, but I thought I would also add that our
Google reps recommended temporarily not using a zone when we hit these
errors.

I'll write up a more detailed reply when I wake up.

soulshake:
> Thanks for the feedback AJ. I'm just about to hit the hay, but I thought I would also add that our Google reps recommended temporarily not using a zone when we hit these errors. I'll write up a more detailed reply when I wake up.


This definitely makes more sense now that I know we had hit `ZONE_RESOURCE_POOL_EXHAUSTED` errors during the last incident. :+1:

backend/gce.go
```diff
@@ -92,10 +91,11 @@ var (
 		"IMAGE_SELECTOR_URL":     "URL for image selector API, used only when image selector is \"api\"",
 		"IMAGE_[ALIAS_]{ALIAS}":  "full name for a given alias given via IMAGE_ALIASES, where the alias form in the key is uppercased and normalized by replacing non-alphanumerics with _",
 		"MACHINE_TYPE":           fmt.Sprintf("machine name (default %q)", defaultGCEMachineType),
+		"MINIMUM_CPU_PLATFORM":   "minimum cpu platform",
 		"NETWORK":                fmt.Sprintf("network name (default %q)", defaultGCENetwork),
 		"PREEMPTIBLE":            "boot job instances with preemptible flag enabled (default false)",
 		"PREMIUM_MACHINE_TYPE":   fmt.Sprintf("premium machine type (default %q)", defaultGCEPremiumMachineType),
-		"PROJECT_ID":             "[REQUIRED] GCE project id",
+		"PROJECT_ID":             "[REQUIRED] GCE project id (will try to auto detect it if not set)",
```
joepvd: In that case, `[REQUIRED]` can be dropped

backend/gce.go
```diff
@@ -119,7 +119,7 @@ var (
 		"WARMER_URL":            "URL for warmer service",
 		"WARMER_TIMEOUT":        fmt.Sprintf("timeout for requests to warmer service (default %v)", defaultGCEWarmerTimeout),
 		"WARMER_SSH_PASSPHRASE": fmt.Sprintf("The passphrase used to decipher instace SSH keys"),
-		"ZONE":                  fmt.Sprintf("zone name (default %q)", defaultGCEZone),
+		"ZONE":                  "zone in which to deploy job instaces into (default is to use all zones in the region)",
```
soulshake: typo (`instaces`)

PR #585, opened 2019-04-09
backend: Add LXD support

stgraber:
Adds a local LXD daemon as a supported backend for the Travis-CI worker.

Took the docker backend, removed most of the code and replaced by the LXD equivalent.

## How can you test this?
No idea :)

@joshk can hopefully take it from here for a bit :)


joshk:
@stgraber AMAZING!

PR #588, opened 2019-05-08
Specifying an unsupported language + `dist: xenial` routes the build to trusty

DamianSzymanski:

There is an issue when user chooses the unsupported language:
https://github.com/travis-ci/bugs-questions-comments-ideas/issues/130

## How can you test this?

The fix is deployed to staging


DamianSzymanski:
> Looks good, but let's combine it with [this](https://github.com/travis-ci/worker/pull/588/files#diff-9a87b0a1b29a3dd92a230acf1c0fb2afR252-R254) under the same `if` to make it more obvious
> 
> ```
> if hasDist {
> 	addTags("dist:" + params.Dist)
>     	addDefaultTag("dist:" + params.Dist)
> }
> ```
@pavel-d if we do that, order will be different, currently default lang is first and the dist

pavel-d:
@DamianSzymanski, got it, makes sense

PR #608, opened 2019-08-14
Don't throw error if machine type self link doesn't exist

DamianSzymanski:
We want to use new gen machine types n2, c2, but they are only available in us-central1-a and us-central1-c, worker was not able to start, failed with an error "failed to find machine type self link"

Remove error if machine type is not present in the zone and throw an error when link map is emtpy.

Any welcome

### DIFF
--- a/backend/gce.go
+++ b/backend/gce.go
@@ -813,13 +813,13 @@ func (p *gceProvider) Setup(ctx gocontext.Context) error {
 					p.machineTypeSelfLinks[key] = mt.SelfLink
 					return nil
 				}
-				return mtErr
+				return nil
 			})
 
-			if err != nil {
-				return errors.Wrap(err, "failed to find machine type self link")
-			}
 		}
+	}
+	if len(p.machineTypeSelfLinks) == 0 {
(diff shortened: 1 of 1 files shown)


backend/gce.go
```diff
@@ -813,14 +813,14 @@ func (p *gceProvider) Setup(ctx gocontext.Context) error {
 					p.machineTypeSelfLinks[key] = mt.SelfLink
 					return nil
 				}
-				return mtErr
```
pavel-d: Not sure how important this error is, maybe we should log it?

backend/gce.go
```diff
@@ -813,14 +813,14 @@ func (p *gceProvider) Setup(ctx gocontext.Context) error {
...
+				return nil
 			})
 
-			if err != nil {
-				return errors.Wrap(err, "failed to find machine type self link")
-			}
 		}
 	}
+	if len(p.machineTypeSelfLinks) == 0 {
+		return errors.Wrap(errors.New("couldn't valid any MachineType"), "built machine type self link map is empty")
```
pavel-d: You meant `couldn't find...` or `couldn't validate...`?

pavel-d:
Closing in favor of https://github.com/travis-ci/worker/pull/621

PR #611, opened 2019-09-16
Services start support on Bionic for TCI Enterprise in docker

JaroslawSokol:
Builds which uses services can not start via systemctl because of some profile permission issue in docker.
`System has not been booted with systemd as init system (PID 1). Can't operate.`

Solution requires to have support to set following envs:
`TRAVIS_WORKER_DOCKER_BINDS`
`TRAVIS_WORKER_DOCKER_SECURITY_OPT`
`TRAVIS_WORKER_DOCKER_TMPFS_MAP`

Issue is related to Travis CI Enterprise docker solution.
Internal ticket:
https://travisci.assembla.com/spaces/Enterprise/tickets/realtime_cardwall?ticket=37

Solution based on information from https://github.com/defn/docker-systemd

## How can you test this?
Install worker with bionic support for Travis CI Enterprise docker solution.
https://docs.travis-ci.com/user/enterprise/setting-up-travis-ci-enterprise/

### DIFF
--- a/backend/docker.go
+++ b/backend/docker.go
@@ -46,14 +46,14 @@ var (
 	defaultDockerSSHDialTimeout                = 5 * time.Second
 	defaultInspectInterval                     = 500 * time.Millisecond
 	defaultExecCmd                             = "bash /home/travis/build.sh"
-	defaultTmpfsMap                            = map[string]string{"/run": "rw,nosuid,nodev,exec,noatime,size=65536k", "/run/lock": "rw,nosuid,nodev,exec,noatime,size=65536k"}
+	defaultTmpfsMap                            = map[string]string{"/run": "rw,nosuid,nodev,exec,noatime,size=65536k"}
 	dockerHelp                                 = map[string]string{
 		"ENDPOINT / HOST":     "[REQUIRED] tcp or unix address for connecting to Docker",
 		"CERT_PATH":           "directory where ca.pem, cert.pem, and key.pem are located (default \"\")",
 		"CMD":                 "command (CMD) to run when creating containers (default \"/sbin/init\")",
 		"EXEC_CMD":            fmt.Sprintf("command to run via exec/ssh (default %q)", defaultExecCmd),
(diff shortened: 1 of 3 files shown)


backend/docker.go
```diff
@@ -53,7 +53,7 @@ var (
 		"CMD":                 "command (CMD) to run when creating containers (default \"/sbin/init\")",
 		"EXEC_CMD":            fmt.Sprintf("command to run via exec/ssh (default %q)", defaultExecCmd),
 		"INSPECT_INTERVAL":    fmt.Sprintf("time to wait between container inspections as duration (default %q)", defaultInspectInterval),
-		"TMPFS_MAP":           fmt.Sprintf("comma- or space-delimited key:value map of tmpfs mounts (default %q)", defaultTmpfsMap),
+		"TMPFS_MAP":           fmt.Sprintf("\"+\"-delimited key:value map of tmpfs mounts (example \"/run:rw,exec+/run/lock:rw,exec\", default %q)", defaultTmpfsMap),
```
pavel-d: @JaroslawSokol, why `+` as a delimiter? is it something standard?

backend/docker.go
```diff
@@ -187,7 +189,12 @@ func newDockerProvider(cfg *config.ProviderConfig) (Provider, error) {
 		binds = strings.Split(cfg.Get("BINDS"), " ")
 	}
 
-	tmpFs := str2map(cfg.Get("TMPFS_MAP"))
+	securityOpt := []string{}
+	if cfg.IsSet("SECURITY_OPT") {
+		securityOpt = strings.Split(cfg.Get("SECURITY_OPT"), " ")
+	}
+
+	tmpFs := str2map(cfg.Get("TMPFS_MAP"), " ")
```
pavel-d: @JaroslawSokol, how is it going to split a string from the example above? `/run:rw,exec+/run/lock:rw,exec`

JaroslawSokol: It is done by config/provider_config.go
System ENV variable like this KEY=VALUE_PART_1+VALUE_PART_2
is finally converted to provider config map seen as
"KEY" with value "VALUE_PART_1 VALUE_PART_2"

PR #617, opened 2019-10-28
Report job error when worker fails to unmarshal job config

pavel-d:
When using invalid types in `.travis.yml` for some keys (e.g. passing `language` as an array, instead of string), worker is unable to unmarshall such config and it silently fails (job gets stuck in queued state). 

Stale jobs get cleaned up after some time by `scheduler`, but that then also reduces the capacity people have available (because `scheduler` thinks the job somehow is still waiting or running)

Worker will report job error back to hub, which should terminate the job correctly and show error in the job logs.

## How can you test this?
Deploy to staging and use invalid `.travis.yml`


amqp_job_queue.go
```diff
@@ -193,11 +194,20 @@ func (q *AMQPJobQueue) Jobs(ctx gocontext.Context) (outChan <-chan Job, err erro
 
 				err = json.Unmarshal(delivery.Body, &startAttrs)
```
pavel-d: We fail to unmarshal `StartAttributes` (generated from user's .travis.yml), but we have successfully unmarshalled the `JobPayload` that contains job ID, so we don't need to extract job ID.

PR #622, opened 2019-11-25
Retry LXC container delete a few times to work around ZFS busy failures

pavel-d:
LXD delete command sometimes fails because of ZFS busy failures

Retry LXD delete few times


pavel-d:
We need to find a different approach

PR #632, opened 2020-04-20
Add an option to define security.privileged

rpsene:

Allow setting security.privileged when running a LXD container via worker.env by setting 
export TRAVIS_WORKER_LXD_SECURITY_PRIVILEGED="true", false is the default.

When running, it will set it properly:
```
config:
  image.description: Travis CI Ubuntu 18.04 Full build env template!
  limits.cpu.allowance: 300%
  limits.memory: 6GB
  limits.processes: "5000"
  linux.kernel_modules: overlay
  security.devlxd: "false"
  security.idmap.isolated: "true"
  security.idmap.size: "100000"
  security.nesting: "true"
  security.privileged: "true"
```

Add a new variable that makes it option, default is false.

## How can you test this?
Set it to true and when doing a Travis build running in a LXD, get the configuration of the running container. This preference should be there.

A review is OK :o)

### DIFF
--- a/backend/lxd.go
+++ b/backend/lxd.go
@@ -41,6 +41,7 @@ var (
 	lxdDockerPool               = ""
 	lxdDockerDisk               = "10GB"
 	lxdNetworkIPv6Filtering     = "true"
+	lxdSecurityPrivileged		= "false"
 
 	lxdHelp = map[string]string{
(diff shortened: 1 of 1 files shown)


rpsene:
@DamianSzymanski ˆ

pavel-d:
@rpsene, built and released

PR #637, opened 2020-08-28
Fix for multiple vm's started by worker

r-arek:
Sometimes worker was starting instance, crashing and then starting the same instance again. 
thanks to @pavel-d

### DIFF
--- a/backend/ec2.go
+++ b/backend/ec2.go
@@ -516,9 +516,6 @@ func (p *ec2Provider) Start(ctx gocontext.Context, startAttributes *StartAttribu
 						instanceChan <- instance
 						return
 					}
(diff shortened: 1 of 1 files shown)


backend/ec2.go
```diff
@@ -516,6 +516,9 @@ func (p *ec2Provider) Start(ctx gocontext.Context, startAttributes *StartAttribu
 						instanceChan <- instance
 						return
 					}
+					if reservation.Instances[0].InstanceId != nil {
+                        logger.Debugf("instance %s is not ready", *reservation.Instances[0].InstanceId)
```
pavel-d, line 520: It has to go just before `time.Sleep(500 * time.Millisecond)`

pavel-d (approved):

PR #651, opened 2021-09-20
vpc: first iteration of implementation

inahga:
This is the first iteration of implementing the IBM Cloud VPC backend. It is capable of spawning a virtual machine in response to a new build request, and establishing SSH connectivity to it.

The rough flow goes:
1. Worker initializes and configures the VPC provider with environment variables (`newVPCProvider()`)
2. Worker recognizes that a build is queued. Worker calls `Start()`
3. Start() does the following:
    1. Generate an SSH key pair and load it into IBM Cloud (`createSSHKey()`)
    2. Create the VM instance (`getInstancePrototype()` and `createInstance()`). It creates it on a random subnet of the ones available.
    3. Wait for the VM instance to become ready and reachable with SSH (`waitForInstance()` and `waitForInstanceSSH()`)
4. Start() returns and Travis moves on to execute the build, which isn't implemented yet so it'll always return success.

The following assumptions are made about the environment:
1. A VPC is created with a set of subnets. The VPC ID and subnet IDs must be provided.
2. The worker has private network access to the VPC where builds are spawned, one way or another. I've been using a VM in the same VPC.

I tested this by running it and ensuring VMs were spawned. I played with the timeouts and made artificial `return err`s to simulate different failures and ensured resources were cleaned up appropriately. The only case not covered if the worker completely dies without cleaning up its resources. In this case we may need to ensure we have a garbage collect script that cleans up the VPC of any old resources. Specific test environment setup is documented on https://github.ibm.com/river/river/issues/685.


inahga:
Sorry! Meant to open this on local fork.

